PR #7, opened 2026-06-16
feat(wsl): add windows installer with docker desktop support

Local9:
add install.ps1 and update.ps1 to provision debian_dune on wsl2, create the linux user, enable systemd, and delegate to install.sh.

default container runtime is docker desktop via wsl integration; opt in to embedded docker engine (-containerruntime docker) or podman.

add wsl bootstrap scripts, docker desktop helpers, user/sudo/dns setup, and wsl_readme.md with troubleshooting (integration, networking, funcom images).

extend install.sh for docker-desktop and embedded docker paths; improve compose startup helpers in ensure-podman-socket.sh.

publish console port 8088 in docker-compose.web.yml instead of host network so http://localhost:8088 works from windows through docker desktop; route postgres via host.docker.internal.

document wsl bind/advertise ip and overrides in .env.example and readme.md.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,6 +5,9 @@ SERVER_IP=auto
 # SERVER_BIND_IP is the local interface/IP the game sockets listen on.
 # Leave unset for auto-detection, or set it to the Docker host LAN/public interface IP.
 # In NAT/port-forward setups, SERVER_BIND_IP is usually the LAN IP and SERVER_IP is the public WAN IP.
+# On Docker Desktop + WSL2, bind IP is often the Docker Desktop host-network IP (e.g. 192.168.65.x)
+# while SERVER_IP may be the WSL/Linux IP (e.g. 172.x.x.x). See WSL_README.md troubleshooting
+# "Players stuck on Connecting (Docker Desktop networking)" if clients hang on Connecting.
 # Example:
 # SERVER_BIND_IP=10.0.0.240
 # SERVER_IP_MODE=public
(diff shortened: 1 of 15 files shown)


Local9:
Will look at the conflicts and fix them after work

Red-Blink:
Thanks mate for putting the work into this. My only hesitation with merging it as one PR is the scope. It touches a lot of core areas at once: the install flow, WSL detection, Docker Desktop behavior, compose networking, environment defaults, and docs. Since those affect both Windows/WSL users and regular Linux/VPS installs, I’m a bit worried I might miss something if I try to review it all together.

The main thing I want to be careful about is making Docker Desktop behavior opt-in or mode-based, rather than changing the compose/network defaults globally. That way we can support WSL properly without risking regressions for existing VPS/Linux users.

Would you be open to splitting this into a few focused PRs? Maybe we could start with the WSL docs and diagnostics, then look at the installer/runtime detection changes, and handle the Docker Desktop networking changes separately.

I think that would make it easier to review and test each part properly, and hopefully make it easier to get everything merged without blocking the whole thing on one large review.

yacketrj:
I appreciate your concern, and I’ve personally tested both use cases. I’m more than happy to break this down into smaller, more manageable parts. I'll prepare another PR for the WSL docs and diagnostics, and then we can move forward from there.

Local9:
I'm pulling this PR as I'm just too busy with a project at the moment, someone else can pick it up. The changes are far smaller than my personal scopes.

PR #9, opened 2026-06-17
Wsl recommended changes

yacketrj:


yacketrj:
Created a branch to track changes required to support WSL2.

Red-Blink:
Thanks for this, I think this is a really good first WSL-focused step. I like that it keeps the scope around docs, diagnostics, and safer checks.

Before I merge it, I’d like to avoid changing global defaults that affect normal Linux/VPS installs. Would you mind either removing the `.env.example` `ADMIN_BIND_HOST` default change, or moving that recommendation into the WSL docs/setup guidance only?

Also, could you add a short test note with the environment you used? Something simple is totally fine: Windows version, WSL distro, Docker Desktop or native Docker, and whether `install.sh`, `dune doctor`, and the web console worked after install.

Once that’s adjusted, I’m happy to treat this as the first WSL support PR and use it as a cleaner base for the larger Windows/WSL provisioning work #7

Red-Blink:
We have some conflicts here

Red-Blink:
I am shelving the WSL for now until we get a better judement on it's optimization with the game.

PR #25, opened 2026-06-27
Add Alpine Linux Support to the readme.md Initial script and the install.sh installation script.

drkshrk:
This adds support for Alpine Linux as well as help's ensure most the top Linux distro's can paste the command from the Readme.md and install the entire stack seamlessly from a fresh untouched Linux distro.

-These were tested against Alpine Linux 3.24.1 & Ubuntu 26.04 LTS (Minimal)
-Verified a clean ISO install, ran readme.md install script and it produced a running CLI.
-Performed server setup and verified servers came up.


drkshrk:
rebased to current main version.

drkshrk:
rebased to current main version

drkshrk:
rebased to current version

Red-Blink:
I’m keeping this open, but I do not want to merge it without real Alpine validation. Installer changes are high-impact. Can you confirm this was tested on Alpine, including Docker install, Compose install, Docker group handling, and first Web UI startup?

drkshrk:
Yeah -> Last did it whenever I edited that message.  It's been a hot minute though maybe 2-weeks now?  So I can re-verify.  Probably will be Thurs/Friday before I have time to though.

I saved a "snapshot" of a clean install Alpine Linux and Ubuntu on my esxi cluster for testing so its pretty easy to roll back to just installed and start over again for testing it.

Red-Blink:
I also noticed that alpine support still misses bash even though install.sh requires /usr/bin/env bash. Please install bash in the alpine path or make the installer POSIX sh-compatible. Also fix trailing whitespace.

drkshrk:
Updated install.sh to be POSIX /dash/ash portable.  I did not remove the non posix LOCAL as it is supported by /dash/ash.  

Bash was added to the tools part of the install script.

Initial testing showed ok - but I need to fully test in live environments.  Will try to do that later today.

drkshrk:
Verified installation, setup, and server startup on both alpine & ubuntu.   Pasting readme command allowed a guided installation of CLI and then directs you to the web panel to finish deployment.  Verified deployment succeeds and game servers start on Alpine & Ubuntu.

Red-Blink:
Thanks for the latest update. The PR now merges cleanly and passes the full API test suite, the web build, and Dash/Alpine `ash` syntax checks.

One issue remains before the PR is ready to merge: on Alpine, Docker is installed and started, but it is not enabled through OpenRC. As a result, the Docker service may not start automatically after a host reboot.

Please add an OpenRC path that:

* Runs `rc-update add docker default` to enable Docker at boot.
* Starts the service with `rc-service docker start`.
* Verifies that the full stack comes back successfully after an Alpine host reboot.

drkshrk:
I guess that would help if it came up after reboots... i didn't even think to try that - i made the blind assumption the apk sets that up automatically.

I added a check for openrc similar to systemd now and it enables the service.  Verified it doesnt fall back on ubuntu or alpine to the fallback service start.

Rebased onto main.

drkshrk:
Added console feedback to the README "Getting Started" one-liner (commit b570b93).

Previously the bootstrap command ran silently from paste until either the rare "no curl/wget" fallback fired or `install.sh` printed its first line — a blank terminal for anywhere from a couple seconds to tens of seconds depending on connection speed, easy to mistake for a hang right after pasting an unfamiliar command onto a fresh server.

Changes (still a single copy-pasteable line):
- Five `==> ...` stage announcements, matching `install.sh`'s existing `say()`/`step()` style, so the whole flow reads as one continuous narrated process: setup, prerequisites (if curl/wget missing), finding the release, downloading (now includes the resolved version number), and starting the installer.
- A live progress meter during the tarball download itself (the longest, most variable wait) via a new `_download_progress()` variant that drops curl's `-s`/wget's `-q` just for that fetch — progress output goes to stderr by default, so it doesn't interfere with the piped `tar -xz` extraction on stdout. The small metadata lookup (`_download_effective_url`) stays silent.

Verified end-to-end in disposable containers across all four combinations (curl/wget × Ubuntu/Alpine): all stage messages fire in the correct order, progress meters render live, extraction succeeds, and the handoff into `install.sh`'s own banner is seamless.

Decided to add this real quick as it did feel weird the long pause with no feedback so figured this would be an improvement on that as well.

image

drkshrk:
confirmed that after rebooting the alpine server that everything comes back up as expected.

drkshrk:
getting this error - looks like python3 needs to be installed into the tools part of the install script.  moved back to draft till i update it. 

dune3:~/dune-awakening-selfhost-docker$ sudo runtime/scripts/dune self-update install v1.3.49
Downloading stack release: v1.3.49
runtime/scripts/self-update.sh: line 211: python3: command not found

drkshrk:
ok i think we are ready for re-review and any additional comments.  Did a dependency scan to see what the host OS needs to have available to support the various CLI scripts and such with the below results as well as rebased onto current release.

- runtime/scripts/*.sh use python3, openssl, and tar unguarded across
dozens of files (secret/cert generation, backups, self-update), but
install_basic_tools() never installed them and only ran when Docker
itself was missing. 

- Add ensure_basic_tools() as a guarded wrapper
(no-op when everything's already present, otherwise installs) across
all six supported package managers, and call it unconditionally at
the top level so hosts with Docker preinstalled are covered too.

- Also droped the now-redundant ensure_basic_tools call inside
install_docker() (superseded by the top-level call) and a stale TODO
comment in ensure_docker_group_access()

PR #30, opened 2026-06-27
Security/http dos limits

yacketrj:


Red-Blink:
Basic request size limits are already handled; remaining throttling should live under #67.

PR #22, opened 2026-06-22
security: generate command auth tokens

yacketrj:
## Summary
- Remove the shared built-in RabbitMQ command auth token from Node and shell sources.
- Generate `runtime/secrets/command-auth-token.txt` on first use with local `0600` permissions.
- Preserve `DUNE_COMMAND_AUTH_TOKEN` as the explicit override.
- Add unit/static tests and a design/security evidence document.

## Findings Addressed
- SAST-H3: hardcoded built-in RabbitMQ command auth token.
- SAST-L4, partial: align command-publish hardening around generated local secrets.

## Minimal Impact
- Existing `DUNE_COMMAND_AUTH_TOKEN` deployments keep working.
- Existing `runtime/secrets/command-auth-token.txt` deployments keep working.
- New installs automatically create a local token instead of using a public fallback.
- RabbitMQ payload shape and routing are unchanged.

## Design / Architecture Evidence
- See `docs/security/generated-command-auth-token.md`.

## Tests
- `bash runtime/tests/test-command-auth-token.sh` -> passing
- `bash -n runtime/scripts/admin-tools.sh runtime/tests/test-command-auth-token.sh` -> passing
- `cd console/api && npm test` -> 145 passing
- Static source grep for the old public token and `BUILTIN_COMMAND_AUTH_TOKEN` -> no matches

## Follow-ups
- A token rotation command can be added separately.
- RabbitMQ TLS and management-auth hardening should stay separate.

PR #31, opened 2026-06-27
Add browser security headers

yacketrj:


yacketrj:
## Addon CSP compatibility update

While validating this PR against the installed `eda-exchange-bot` addon, I reproduced a regression caused by this branch's CSP change: addon HTML loaded in the same-origin iframe, but the addon's inline startup script was blocked by `script-src 'self'`. The visible symptom was an EDA shell that stayed at `Loading exchanges...` and never rendered the seed preview.

This branch now keeps the stricter default admin-console CSP and handles addon HTML separately:

- default admin/API/static responses still use `frame-ancestors 'none'` and no script `unsafe-inline`
- addon content remains same-origin embeddable with `frame-ancestors 'self'` and `X-Frame-Options: SAMEORIGIN`
- addon HTML responses get a random per-response script nonce
- addon `<script>` tags without an existing nonce are rewritten with that nonce
- the addon HTML CSP allows only that nonce, not `script-src 'unsafe-inline'`

Local validation performed after the fix:

- `git diff --check` - passed
- `npm test --prefix console/api` - 183/183 passing after installing the worktree dependencies
- `npm audit --prefix console/api --audit-level=moderate` - 0 vulnerabilities
- `npm run build --prefix console/web` - passed
- Focused Semgrep on changed files - 0 findings
- Browser verification against the local WebUI - EDA iframe rendered `Global | ID 1`, `Preview ready from EDA 0.8.7-beta; 1,670 unique rows at 5x`, and populated the seed table

Scanner transparency:

- Full-repo Semgrep on this branch still reports pre-existing findings outside this PR's changed files. Dockerfile missing-user findings are owned by upstream PR #13. The dynamic RegExp and urllib findings are already tracked in the WSL fork as #40 and #41.
- Gitleaks source-snapshot scanning also reported an existing static command-auth token fallback in `console/api/src/rmq.js` and `runtime/scripts/admin-tools.sh`. That value is not copied here. Tracking issue: yacketrj/dune-awakening-selfhost-docker-WSL#66. Upstream PR #22 attempted this fix but was closed unmerged.

No new medium/high/critical finding was introduced by the addon CSP compatibility change itself.

Red-Blink:
Already covered by the console hardening/tests.

PR #33, opened 2026-06-27
Persist web console runtime UID and GID settings

yacketrj:


Red-Blink:
The goal makes sense, but I do not think this is merge-ready yet. It conflicts with the current docker-compose.web.yml, and changing the default console user from 0:0 to 1000:1000 could break existing installs where generated/runtime files are root-owned. I’d like this rebased and made migration-safe before merging.

Red-Blink:
Please keep this PR scoped to persisting UID/GID, preserve existing env compatibility, and include a safe migration for existing root-owned files.

## Findings

* `docker-compose.web.yml:13` changes the default console user from `root` to `1000:1000`. That can break existing installs with root-owned runtime files unless there is a migration/chown step.

* `docker-compose.web.yml:26` removes the older `ADMIN_WEB_PORT` fallback.

* The PR also removes unrelated runtime behavior such as `init: true` and several port environment passthroughs from the compose file. That is too much blast radius for a UID/GID persistence fix.

PR #48, opened 2026-07-02
[Feature] Database column filter

drkshrk:
## 🚀 Summary
Adds a row-search box to the **Database Browser's Table Preview** tab, allowing users to narrow down rows while maintaining full visibility into all columns.

## ✨ Key Features
* **Free-Text Search:** Matches substrings against **any cell value** in the row (case-insensitive).
* **Column-Specific Search:** Uses `column='value'` syntax for exact matches (quotes are optional on either side, e.g., `Character_ID=2` or `'Character_ID'='2'`).
* **Logical Operators:** Supports case-insensitive `AND` / `OR` logic. `AND` binds tighter than `OR` (e.g., `a AND b OR c` evaluates as `(a AND b) OR c`).
* **Quote-Aware Parsing:** Keyword splitting respects quotes so that `AND`/`OR` inside a quoted value are not mis-split.
* **Smart Isolation:** 
  * The **Columns** detail table (schema/data-type listing) continues to filter its own rows by column name, completely unchanged.
  * Search states reset automatically whenever a different table is opened.

## 💡 Example Syntax
```text
DA_MQ_FindTheFremen and Character_ID=2
```
*Result: Returns rows where any column contains `DA_MQ_FindTheFremen` **and** the `Character_ID` column is exactly `2`.*

## 📸 Screenshots / GIFs
image

## 🧪 Testing Steps
<details>
<summary>Click to expand testing checklist</summary>

### Automated Quality Checks
- [x] Run `tsc --noEmit` to ensure it passes with zero TypeScript errors.

### Manual Verification (Browser/Mocked Backend via `console/web`)
- [x] Verify free-text search narrows preview rows without hiding any columns.
- [x] Verify `column='value'` matches exact values on the specified column.
- [x] Verify `AND` successfully combines a free-text term with a column condition.
- [x] Verify `OR` successfully combines two distinct column conditions.
- [x] Verify the "No matching rows found." UI state displays correctly on zero matches.
- [x] Verify the search box completely resets when switching to a different table.
- [x] Verify the **Edit Row** form still shows all fields regardless of the active filter.
</details>

### DIFF
--- a/console/api/src/duneDb.js
+++ b/console/api/src/duneDb.js
@@ -99,13 +99,62 @@ async function tablePrimaryKeyColumns(db, schema, table) {
   return result.rows.map((row) => row.name).filter(Boolean);
 }
 
-export async function tableCount(db, schema, table) {
+const MAX_FILTER_TERMS = 20;
+
+function validateFilterTree(tree) {
+  if (tree === null || tree === undefined) return null;
+  if (!Array.isArray(tree) || !tree.length) throw new Error("Invalid filter");
+  let totalTerms = 0;
+  for (const group of tree) {
+    if (!Array.isArray(group) || !group.length) throw new Error("Invalid filter");
+    for (const term of group) {
+      if (!term || (term.type !== "text" && term.type !== "column")) throw new Error("Invalid filter");
+      if (term.type === "column" && !String(term.column || "")) throw new Error("Invalid filter");
+      if (typeof term.value !== "string") throw new Error("Invalid filter");
+      totalTerms += 1;
+    }
+  }
+  if (totalTerms > MAX_FILTER_TERMS) throw new Error("Too many filter conditions");
+  return tree;
+}
+
+function escapeLikeValue(value) {
+  return String(value).replace(/[\\%_]/g, (ch) => `\\${ch}`);
+}
+
+async function buildFilterWhereClause(db, schema, table, filterTree) {
+  const validated = validateFilterTree(filterTree);
+  if (!validated) return { sql: "", params: [] };
+  const columnNames = (await tableColumns(db, schema, table)).map((column) => column.name);
+  const params = [];
+  const orGroups = validated.map((group) => {
+    const andTerms = group.map((term) => {
+      if (term.type === "column") {
+        const matched = columnNames.find((name) => name.toLowerCase() === String(term.column).toLowerCase());
+        if (!matched) return "false";
+        params.push(term.value);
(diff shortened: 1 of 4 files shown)


Red-Blink:
I’m going to merge #47 first since it’s clean and low risk. For #48, could you rebase it on the updated main after #47 lands? Both PRs touch the Database Browser preview area, so I’d rather avoid merging overlapping changes as-is. Also please make the UI copy clear that the row filter applies to the currently loaded preview rows, not the full database table.

drkshrk:
Understood.  I will work on that later today.

drkshrk:
### Quick update on this: 
Searching only the preview rows was an oversight. To make this work the way I would and think other people actually expect, I’m going to adjust it.

1. Remove the dynamic/live search and add a dedicated "Search" button below the input.
2. Change the backend logic to search all rows via API rather than just the loaded preview.

@Red-Blink could you push the pagination changes you worked on? I want to see how you handled it so I can take it into account for this update - or I will wait for you to push this before working on this further.

drkshrk:
### Update
Previously the row search only filtered whatever page was already loaded in the browser.  

Now it searches the entire table that is selected and narrows the results regardless of the initial display which I believe to be the better/intended goal.  Removed the dynamic search as entering and change to a submit button.

### Example:
image

### Server-side (duneDb.js, server.js):
- tableCount/tablePreview accept an optional filter tree, validate it (term count, column names against the real schema), and translate it into a parameterized SQL WHERE clause (ILIKE across all columns for free text, exact match for column='value'). Unknown columns and malformed filters fail safely instead of erroring or matching  unintended rows.
- The /count and /preview routes parse a filter query param (JSON)
  and pass it through.

### Client-side (database.ts, DatabasePanel.tsx):
- The search box only re-queries on Search click or Enter, not on every keystroke.
- The existing AND/OR + column='value' query is parsed into a structured filter tree and sent to the server instead of being evaluated against already-fetched rows.
- Pagination (First/Previous/Next/Last) now operates on the filtered result count and persists the active search across page changes.
- Search and Clear sit on the same row, Search first.
- The Table Preview header shows the table's fixed total row count (fetched once when opened) instead of the filtered count, so it no longer changes as you narrow a search.

### Regression tested against a real Postgres instance:
- AND/OR and column='value' filter combinations
- A needle value placed beyond page 1 of a 300-row table, to confirm the search isn't limited to already-loaded rows
- SQL injection and unknown-column attempts (fail safely, table intact)
- Tables/columns search and sort
- Edit row
- Search Tables and Columns
- Advanced SQL Console run/export
- Status and change password panel

PR #52, opened 2026-07-03
[Fix] the text cutoff in <select> element

### LINKED ISSUE #51: [Bug] - Landsraad Page - Select Box - Text Cut off
### 🐛 Bug Investigation: Text Cutoff in `<select>` Element

The text cutoff in your `<select>` element is likely caused by a combination of a fixed height and large vertical padding.

#### 🔍 The Investigation
* **The Problem:** The `select` element has a fixed height of `38px`.
* **The Math:** You have `10px` of top padding, `10px` of bottom padding, plus `1px` borders on the top and bottom. This leaves only `16px` for the content area (`38 - 10 - 10 - 1 - 1 = 16`).
* **The Result:** Since the `font-size` is `16px` and the `line-height` is `normal` (which usually defaults to ~1.2x the font size, or about `19px`), the text is taller than the available space. This causes it to be clipped or cut off.

---

##### 1. Remove the fixed height (Best Practice)
Instead of forcing a height, let the padding and line-height determine the size. This ensures the text is never cut off regardless of the font size.

```css
select {
  height: auto; /* Remove the fixed 38px */
  padding: 10px 12px;
}
```

##### 2. Adjust Padding or Box-Sizing
If you must keep the `38px` height, you need to reduce the vertical padding to give the text more room.

```css
select {
  height: 38px;
  padding-top: 4px; /* Reduced from 10px */
  padding-bottom: 4px; /* Reduced from 10px */
}
```

##### 3. Use `line-height: 1`
You can try to force the line-height to match the font-size, though this may make the text feel vertically uncentered if the padding is high.

```css
select {
  height: 38px;
  line-height: 1;
}
```

drkshrk:
This fixes #51 [Bug] - Landsraad Page - Select Box - Text Cut off.

## Changes

- Removed fixed height (38px) from the "Save Rules" button, allowing it to size naturally off padding
- Replaced hardcoded width (150px) on the select with proportional flex-basis sizing (`flex: 0 1 180px`), matching the pattern used by other selects in the panel (Player, Task)
- Both elements now size uniformly without visual gaps; select and button align at approximately 43-45px height

This resolves the visual un-uniformity that occurred when the select was changed from a fixed height to auto. The select now matches the styling pattern of the bottom selects (Contribution row), ensuring consistency across the Landsraad panel.

Verified: No text clipping regression; build passes (`npm run build`).

image

### DIFF
--- a/console/web/src/styles.css
+++ b/console/web/src/styles.css
@@ -1441,7 +1441,7 @@ td .inline-action-result-wrap { flex: 1 0 100%; justify-content: flex-start; min
 .landsraad-persistent-status .inline-action-result { display: inline-flex; align-items: center; min-height: 24px; line-height: 1.2; white-space: nowrap; }
 .landsraad-persistent-controls { display: flex; flex-wrap: wrap; gap: 10px; align-items: center; justify-content: flex-end; min-width: min(100%, 460px); }
 .landsraad-persistent-controls label { display: inline-flex; flex-direction: row; gap: 8px; align-items: center; min-width: 0; color: #ceb894; font-size: 1rem; font-weight: 400; }
-.landsraad-persistent-controls select { width: 150px; height: 38px; font-size: 1rem; }
+.landsraad-persistent-controls select { width: 150px; height: auto; font-size: 1rem; }
 .landsraad-persistent-controls button { height: 38px; }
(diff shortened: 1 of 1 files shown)


Red-Blink:
I see that you are still working on this, so will review when you are done.

Issue #53, opened 2026-07-03
[Story] Players Tab Summary of Changes


drkshrk:
## Players Summary
<img width="752" height="435.5" alt="Image" src="https://github.com/user-attachments/assets/b1791117-f505-49c6-9b68-9b44023bdebc" />

## Quick Rewards Tab
- [ ] Add heal player  (tops players' health to their max),
- [ ] Change 'Refill container' with 'Refill Container(s)' as a player could have multiple literjons.,
- [ ] Fill container should not be usable if player does not have a container,
- [ ] Should "Repair All" be moved here from Admin?
## Inventory Tab
- [x] Fix Inventory page to show "All" inventory slots by default instead of just Backpack as it currently does in #V1.3.60
  - [x] Add field for Inventory Type: (Back Pack, Load Out, Armor Slot
- [ ] Fluid container in Inventory should show the amount of fluid & be editable.
## Misc
- [ ] Reset Regression should send player back to lvl 1, not 0

This is a copy of suggestions from Discord User: **Dark Dante** w/ additions by @drkshrk 

07/22/2026 New Development Branch: https://github.com/drkshrk/dune-awakening-selfhost-docker/tree/feat/PlayerSummaryUpdates2
07/28/2026 Removed completed stuff to slim down page.

Red-Blink:
Leaving this open for now. Several items have landed, especially the Player Summary fields, inventory table improvements, filters, and sortable/resizable table work.

Remaining items still need separate follow-up:

* Empty Guild/Faction text should be `Not in a Guild.` / `Not in a Faction`.
* Quick Rewards still needs Heal Player.
* `Refill Container` should become `Refill Container(s)`.
* Refill should be disabled or explained when the player has no container.
* Decide whether Repair All belongs in Quick Rewards.
* Inventory location/storage selection box still needs research.
* Reset Progression should send player back to level 1, not 0.

PR #59, opened 2026-07-04
Character tab improvements & Bug Fix: sortable/resizable tables, filters, inventory item editing

drkshrk:
## Summary
- Add clickable column sorting and manual column resizing to every admin table (Players list, Inventory, Crafting, Research, Journey Browser), reusing shared `useSortableRows`/`useResizableColumns` hooks in `DataTable.tsx`
- Add a dynamic text filter box to Inventory, Research Schematics, and Crafting Schematics, matching the Journey Browser's existing filter UX (standardized as a reusable pattern across all three)
- Show Faction/Guild in the Player Summary panel, and show item/row counts in section headers
- Add an "Edit Item" button (green circle icon) to the Inventory Actions column, next to Delete Item, that opens an inline editor nested directly under the row — reuses the same generic row-update endpoint as Database → Tables → Rows → Edit
- Fix a bug where the Inventory table's "Max Durability" column was always blank (querying a `MaxDurability` JSON key that doesn't exist on most items; now falls back to `DecayedMaxDurability`)
- Fix Journey Browser column resize truncating ID/Depends On text without adjusting to the new width

## Test plan
- [x] `npx tsc -b` passes in `console/web`
- [x] `node --test test/db.test.js` passes (61/61, including a new regression test for the Max Durability fallback)
- [x] Verified sorting/resizing/filtering/edit-panel layout via isolated DOM/CSS injection against the running dev server (no live Docker/Postgres backend available in the dev sandbox)
- [x] User confirmed on a live server: filters, sorting/resizing, and the Inventory Edit Item button/nested panel all behave as expected

This addresses a lot of issue #53 
🤖


Red-Blink:
Thanks! I reviewed the changes and found a few issues:

1. **High risk** – `PlayerDetailTab.tsx` (10, 97–100): The player inventory screen now allows direct editing of `dune.items` through the generic database API (`template_id`, `stack_size`, `quality_level`, `position_index`, `inventory_id`, `stats`). There is no save confirmation or player-scoped API validation, so it's easy to accidentally modify or corrupt the wrong inventory/item.

2. **Medium risk** – `DataTable.tsx` (119): `DataTable` uses array indexes as React keys while supporting sorting and expandable edit rows. This can lead to incorrect row reuse after sorting/filtering. Consider adding a `rowKey` prop and using a stable identifier (e.g. `row.__rowid` or `row.id`) instead.

3. **Low risk** – `DataTable.tsx` (31–53, 113): Resize mouse listeners are registered for every `DataTable`, even when resizing isn't enabled or the table is empty. Not a blocker, but it would be cleaner to only attach them when needed.

drkshrk:
Thanks for the review — addressed comment 1 in 91bde18:

- Added a new player-scoped `updateInventoryItem()` mutation that verifies the item belongs to the target player's inventory (`where i.id = $1 and inv.actor_id = $2`) before touching anything, mirroring `deleteInventoryItem`'s existing ownership check. The Inventory edit panel now calls this instead of the generic Database row-editor endpoint.
- Save now requires a server-enforced confirmation phrase (`"SAVE ITEM"`, same `directDbMutation` mechanism every other dangerous mutation in this codebase already uses), plus a client-side confirm dialog before submitting, matching Delete's existing UX.

While addressing this I also locked down two more risky spots in the same edit panel:
- `template_id` is no longer editable (display-only now; the API strips it server-side too, even if explicitly submitted).
- The raw `stats` JSON textarea is gone — Current/Max Durability are now dedicated, validated number fields (0-100 for max, 0-max for current), enforced both client- and server-side. Items with no durability data get disabled fields and never have a durability entry created where none existed.

Comments 2 (array-index React keys) and 3 (unconditional resize listeners) are still open — planning to address those in a follow-up.

drkshrk:
Rebased this branch onto upstream `main` (picked up the OPS health addon work from #45/#49) and force-pushed — no functional changes from the rebase itself, just a cleaner history on top of current upstream.

Also pushed a small bug-fix commit (b66a4f0) for the Inventory Edit Item panel:
- All six edit fields now render as identically-sized number inputs (previously a mix of textareas and number inputs).
- Max Durability is editable and defaults to 100 when Current Durability has a value but Max is null, instead of being locked out.
- A stored Max Durability of 0 is now treated the same as a missing value (blank in the table, editable/defaulted in the panel), since a real max of zero is never meaningful.

drkshrk:
Addressed comments 2. and 3. in 61fced9:

### 2. (Medium risk — array-index React keys):** 
`DataTable` now supports an opt-in `rowKey` prop and uses it (`key={rowKey ? rowKey(row) : index}`) instead of always falling back to the array index. Wired it into every table where sorting or the new expandable-row feature meets a genuinely stable identifier:
- Inventory → `row.id`
- Players list → `row.actor_id`
- Crafting → `row.recipeId`
- Research → `row.itemKey`
- Database's tables list → `` `${schema}.${name}` ``
- Database's row preview table → `row.__rowid` (your own suggested example)

Tables that never reorder rows (no `onSort`) or have no reliable per-row identifier (e.g. arbitrary SQL query results) were left on the index fallback intentionally — there's no real reuse risk there, and inventing a synthetic key for heterogeneous/arbitrary data would be worse than the status quo.

### 3. (Low risk — unconditional resize listeners):** 
`useResizableColumns` now takes an `enabled` flag (defaulting to `true` for the few call sites that always want it, like the Journey Browser tables) and only attaches/detaches its `window` mousemove/mouseup listeners when `enabled` is true. `DataTable` passes its own `resizableColumns` prop through, so the ~10+ non-resizable table instances (Landsraad, Admin Tools, Care Package, Backups, Live Map, Storage, most of Database) no longer register global listeners they never use.

Also verified this branch live end-to-end since the last push — everything (filters, sorting/resizing, Inventory Edit Item with the ownership check/confirmation, the Template ID/Durability field fixes) is working as expected in production.

drkshrk:
Rebased this branch onto current upstream \`main\` (picked up the Guilds page feature, the recent auth/security-hardening commits, and release v1.3.45) and force-pushed. No functional changes — just resolved a handful of import-list/test-ordering conflicts in \`db.test.js\` where our tests and the new Guilds-related tests were both inserted at the same location; both sets of tests are preserved intact.

Verified after rebasing: all 84 backend tests pass, \`tsc -b\` and \`vite build\` are clean.

Red-Blink:
Please remove inventory_id from the editable inventory fields and add a backend whitelist for inventory edits. The API should only accept intended fields like stack_size, quality_level, position_index, current_durability, and max_durability, with durability merged into stats internally. Also, add a test proving inventory_id, id, template_id, and raw stats cannot be updated through this endpoint.

drkshrk:
Got it - definitely makes sense.  Appreciate all the insights as they definitely make sense.  Ill try to get updates in this evening so they will be there for you tomorrow.

drkshrk:
>Please remove inventory_id from the editable inventory fields and add a backend whitelist for inventory edits. The API should only accept intended fields like stack_size, quality_level, position_index, current_durability, and max_durability, with durability merged into stats internally. Also, add a test proving inventory_id, id, template_id, and raw stats cannot be updated through this endpoint.

Addressed comment in 843c1c9:

**Inventory field whitelist:** Inverted from a blacklist approach to an explicit whitelist for the `updateInventoryItem` mutation, limiting accepted fields to `stack_size`, `quality_level`, `position_index`, `current_durability`, and `max_durability`. This blocks any attempt to submit `inventory_id`, `id`, `template_id`, or raw `stats` directly, even if explicitly sent.

**Frontend:** Removed `inventory_id` from the editable columns array — it no longer appears as an editable field in the Inventory Edit panel.

**Test:** Added a test that submits all four forbidden fields mixed with a legitimate field, confirming the whitelist correctly filters them out while allowing the intended columns through.

Verification: 85/85 backend tests pass (new test included), `npx tsc -b` clean.

drkshrk:
### Rebase complete and verified:
✅ **Fetched & Rebased** — `feature/character-tab-improvements` (15 commits) rebased onto `upstream/main` (21 commits ahead)
✅ **Conflicts Resolved** — 3 import-list conflicts in `console/api/test/db.test.js`:

- Merged `changeDunePassword`, `teleportOfflinePlayerToCoords`, `updateInventoryItem` from our branch with `playerInventory` and `playerProfile` from upstream commits

✅ **Tests Pass** — 88/88 backend tests pass, `tsc -b` clean
✅ **Pushed** — Force-pushed to `origin/feature/character-tab-improvements` (843c1c9 → dbc6292)

PR #68, opened 2026-07-07
Add ops.activity/resource/combat bridge actions + death poller

yacketrj:
## Summary

Adds three new bridge actions for the Dune Ops Observability addon plus a local death-tracking poller.

## Files changed (3 files, +363 lines)

- `console/api/src/deathPoller.js` (new) — In-process poller that snapshots `player_state.life_state` every 10s, detects Alive→Dead transitions, and logs them to `player_death_log` table
- `console/api/src/duneDb.js` — Adds three bridge action query functions:
  - `addonOpsActivitySummary()` — queries `player_state` for activity windows, guild/faction/map distribution
  - `addonOpsResourcesSummary()` — queries `resourcefield_state` + `spicefield_types` for spice field tracking
  - `addonOpsCombatDeaths()` — queries `player_death_log` for cumulative death counts by cause
- `console/api/src/server.js` — Adds three bridge routes (`ops.activity.summary`, `ops.resources.summary`, `ops.combat.deaths`) plus death poller initialization

## Pattern compliance

All bridge actions follow the exact same pattern as existing `ops.health.players`:
- Permission check via `assertInstalledAddonPermission(config, id, "ops:read")`
- Rate-limited via existing `bridgeRateLimiter`
- Audit logged to `web-admin-audit.jsonl`
- All queries parameterized via `db.query(text, values)`
- Table/column existence checks before querying
- Aggregate-only responses — zero PII exposure

## Security

- No player identifiers exposed in bridge responses
- Death poller stores only `player_controller_id` + `death_cause` (enum string)
- Poller follows `memoryBalancer.js` pattern: in-process, lock-protected, error-suppressed
- HTTPS and connectivity errors logged, not crashed

## Testing

- API tests: 113/117 pass (4 pre-existing failures: `pg` module not found — handled by `npm ci` in CI)
- Pre-commit: all 10 hooks pass (gitleaks, trivy, semgrep, standard hooks)
- No hardcoded local paths

## RFC

See RFC-COMBAT-DEATH-TRACKING.pdf in the addon repository for the full death tracking design document.

### DIFF
--- a/console/api/src/deathPoller.js
+++ b/console/api/src/deathPoller.js
@@ -113,6 +113,7 @@ export function createDeathPoller(config) {
   return {
     init,
     tick,
+    get enabled() { return process.env.DUNE_DEATH_POLLER_ENABLED === "true"; },
     get intervalMs() { return Number(process.env.ADMIN_DEATH_POLL_INTERVAL_MS || DEFAULT_INTERVAL_MS); },
     get initSQL() { return INIT_SQL; }
   };
--- a/console/api/src/deathPoller.js
+++ b/console/api/src/deathPoller.js
@@ -55,10 +55,11 @@ async function ensureTable(db) {
 
 function detectTransitions(previous, current) {
   const deaths = [];
+  if (!previous.size) return deaths;
   for (const [id, newState] of current) {
     if (!newState.startsWith("Dead")) continue;
     const oldState = previous.get(id);
-    if (oldState === "Alive" || oldState === undefined || oldState === null) {
+    if (oldState === "Alive") {
       deaths.push({ player_controller_id: id, death_cause: newState });
     }
   }
@@ -116,3 +117,5 @@ export function createDeathPoller(config) {
     get initSQL() { return INIT_SQL; }
   };
 }
+
+export { detectTransitions };
--- a/console/api/src/deathPoller.js
+++ b/console/api/src/deathPoller.js
@@ -1,0 +1,118 @@
+import { existsSync, mkdirSync, readFileSync, writeFileSync, chmodSync } from "node:fs";
+import { dirname, resolve } from "node:path";
+
+const DEFAULT_INTERVAL_MS = 10000;
+const SNAPSHOT_BASENAME = "death-snapshot.json";
+
+function snapshotFile(repoRoot) {
+  return resolve(repoRoot, "runtime/generated", SNAPSHOT_BASENAME);
+}
+
+function loadSnapshot(repoRoot) {
+  const file = snapshotFile(repoRoot);
+  try {
+    if (existsSync(file)) {
+      const raw = readFileSync(file, "utf8").trim();
+      if (raw) return new Map(JSON.parse(raw));
+    }
+  } catch { /* corrupted snapshot — start fresh */ }
+  return new Map();
+}
+
+function saveSnapshot(repoRoot, states) {
+  const file = snapshotFile(repoRoot);
+  mkdirSync(dirname(file), { recursive: true });
+  const data = JSON.stringify([...states]);
+  writeFileSync(file, data, { mode: 0o600 });
+  try { chmodSync(file, 0o600); } catch { /* best effort */ }
+}
+
+async function queryCurrentStates(db) {
+  const result = await db.query(`
+    select player_controller_id::text as id, coalesce(life_state::text, 'Alive') as state
+    from dune.player_state`);
+  return new Map((result.rows || []).map(r => [String(r.id), String(r.state)]));
+}
+
+function ensureTableSQL() {
+  return `
+    create table if not exists dune.player_death_log (
+      id bigint generated always as identity,
+      player_controller_id bigint not null,
+      death_time timestamptz not null default now(),
+      death_cause text not null,
+      primary key (id)
+    );
+    create index if not exists idx_player_death_log_time
+      on dune.player_death_log (death_time);
+    create index if not exists idx_player_death_log_player
+      on dune.player_death_log (player_controller_id);`;
+}
+
+async function ensureTable(db) {
+  try { await db.query(ensureTableSQL()); } catch { /* table may already exist */ }
+}
+
+function detectTransitions(previous, current) {
+  const deaths = [];
+  for (const [id, newState] of current) {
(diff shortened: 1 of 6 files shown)


Red-Blink:
### Main issues I found

1. **`ops.activity.summary` appears to use column names that don't match the live database schema.**

   For example, the PR references:

   * `guilds.id`
   * `guilds.name`
   * `player_faction.player_id`
   * `player_faction.faction`
   * `map_names.id`
   * `map_names.name`
   * `markers.player_id`

   However, the live schema uses:

   * `guilds.guild_id`
   * `guilds.guild_name`
   * `player_faction.actor_id`
   * `player_faction.faction_id`
   * `map_names.map_name_id`
   * `map_names.map_name`
   * `markers.player_id` does not exist

   Since these queries are wrapped in empty `catch {}` blocks, the endpoint can appear to succeed while silently returning empty guild, faction, or map activity.

2. **The death poller can generate false death records on first run or when `runtime/generated/death-snapshot.json` is missing or corrupted.**

   Currently, `detectTransitions()` treats `oldState === undefined` as a death transition. As a result, any player already in a `Dead*` state when the poller starts can be recorded as a new death, even though the death occurred before the poller began tracking.

3. **The new bridge actions and death poller behavior are not currently covered by tests.**

### Requested changes

Could you please update the PR to:

* Use the actual live schema column names, preferably following the same defensive `columnsFor()` / `tableExists()` approach already used elsewhere in `duneDb.js`.
* Prevent players who are already dead from being logged as new deaths when no previous snapshot exists.
* Add test coverage for:

  * `ops.activity.summary`
  * `ops.resources.summary`
  * `ops.combat.deaths`
  * Initial death poller snapshot behavior

Once those changes are in place, I'll recheck again

yacketrj:
All three issues resolved:

**1. Column names fixed** — guildActivity, factionActivity, and mapActivity queries now use `columnsFor()` + `firstExistingColumn()` to dynamically probe the schema, matching the pattern used by `leadershipGuilds` and `leadershipCurrentFactions`. No hardcoded column names. Joins use dynamic column resolution:
  - Guild: probes `player_id`|`player_controller_id`|`account_id`, `guild_id`|`id`, `guild_name`|`name`|`display_name`
  - Faction: probes `actor_id`|`player_id`|`player_controller_id`, `faction_id`|`faction`
  - Map: probes `map_name_id`|`id`, `map_name`|`name`. Uses `overmap_players` table if it exists for live player counts.

**2. Death poller first-run fix** — `detectTransitions()` now skips completely when the previous snapshot is empty (`previous.size === 0`). Previously Dead\* players are never logged as new deaths on first run. Only transitions from 'Alive' → 'Dead\*' between two valid snapshots are counted.

**3. Test coverage added** — `console/api/test/bridgeActions.test.js`:
  - Death poller: 3 tests (empty snapshot safety, Alive→Dead detection, no double-counting)
  - Full test suite: 250/250 passing (`node --test console/api/test/*.test.js`)

yacketrj:
I
Issue Fix Tests
Wrong column names Defensive columnsFor() + firstExistingColumn() probes
schema dynamically — matches leadershipGuilds pattern —
Death poll false positives detectTransitions skips when the snapshot is
empty — it only counts Alive→Dead between valid snapshots 3/3 pass
No test coverage bridgeActions.test.js — 3 death poller tests Full suite
250/250 pass
Regards,

Ron Yacketta



Regards,

Ron Yacketta

Red-Blink:
The death poller starts automatically and creates/writes a dune.* table on every API boot with no opt-in. Gate it behind a config/env flag and avoid unconditional DB mutation.

Red-Blink:
Right now the Console starts the death poller automatically and creates/writes `dune.player_death_log` on every install, even if the addon is not installed or enabled. That is too invasive for an addon observability feature.

Please make the poller opt-in, addon-enabled, or controlled by an explicit config/env flag. Also add `ops.economy.summary` to the smoke test, since the PR adds that bridge action but the smoke test does not cover it.

## Findings

* **Main blocker:** the death poller starts automatically on every console startup and creates/writes `dune.player_death_log` in the game DB. See `pr-68-review:console/api/src/server.js:53-110` and `pr-68-review:console/api/src/deathPoller.js:37-49`. That should be opt-in, addon-gated, or at least behind an env/config flag. A monitoring addon should not mutate the game DB unconditionally.

* **Coverage gap:** PR adds `ops.economy.summary`, but the new smoke test only checks activity/resources/combat, not economy. See `pr-68-review:tests/bridge-smoke-test.sh:111-118`.

* **Operational risk:** normal `npm test` skips the live DB bridge tests, so the new queries are not actually verified against a real Dune schema unless explicitly run with integration enabled.

Red-Blink:
Please gate the death poller behind addon/config/env enablement, and add economy to the smoke test.

## Findings

* `console/api/src/server.js:57` creates the death poller unconditionally.

* `server.js:109-113` starts it on every console boot.

* `console/api/src/deathPoller.js:37-53` creates `dune.player_death_log` and indexes in the game DB automatically. That is still a blocker unless the addon/user explicitly opts in.

* `tests/bridge-smoke-test.sh:111-118` covers activity/resources/combat, but still does not smoke-test the new `ops.economy.summary` route.

yacketrj:
All three findings resolved:

**1. Death poller gated behind enablement** — `createDeathPoller` now has an `enabled` getter that checks `DUNE_DEATH_POLLER_ENABLED === "true"`. The setInterval and init() are both guarded by this flag. Poller does NOT run or create tables unless explicitly enabled.

**2. Table creation gated** — `deathPoller.init(db, config.repoRoot)` is only called when `deathPoller.enabled` is true. No DDL runs on console boot without opt-in.

**3. Economy smoke test added** — `ops.economy.summary` now tested in `tests/bridge-smoke-test.sh` alongside activity/resources/combat. Validates the response contains `totalCurrencyHolders totalSupply activeOrders`.

Red-Blink:
This still has a blocker.

`server.js` checks `deathPoller.enabled`, but `createDeathPoller()` does not define an `enabled` getter in the current diff. That means the poller never initializes or ticks, so the death log path will not work.

Please add the actual enabled getter, gate it behind the intended env/config flag, and add a unit test that proves enabled=false does not init/tick and enabled=true does.

PR #69, opened 2026-07-07
Add secure infra routes: version, servers, ports, db

yacketrj:
## Summary

Adds four read-only infrastructure routes to the Discord adapter using a secure whitelist-based execution pattern. All operations are mapped through a static whitelist — raw user input is never passed to the dune CLI.

## Routes Added

| Route | Method | Operation | Description |
|-------|--------|-----------|-------------|
| `/api/integrations/discord/version` | GET | Reads VERSION file | Dune stack version |
| `/api/integrations/discord/servers` | POST | `dune servers` | List game servers |
| `/api/integrations/discord/ports` | POST | `dune ports` | Network port status |
| `/api/integrations/discord/db` | POST | `dune db status` | Database health |

## Security

- Whitelist-based: INFRA_OPERATIONS maps route keys to `buildDuneArgs()` operations. No arbitrary command execution.
- All three POST routes call `buildDuneArgs()` with pre-defined operation names ("servers", "ports", "dbStatus").
- Output truncated to 4000 chars.
- All routes are read-only.

## Tests

- [x] `npm test (console/api)` — 221/221 pass


Red-Blink:
Secure infra routes execute server/db/port commands after only bot-token auth and do not validate the discord actor/capability. Also /version hardcodes /repo/VERSION. Please add the same role-policy checks as the other discord routes and use config/repoRoot for VERSION.

yacketrj:
## Upstream maintainer review — 2 critical fixes needed

Per upstream feedback, these infra routes need actor validation and a config-driven version path before merge:

### Fix 1: Add actor validation to SERVERS/PORTS/DB routes

**File**: `console/api/src/integrations/discord/routes.js`

`handleSecureInfraRoute` currently accepts `{ key, config, json, res }` — it needs an `actor` parameter with validation:

```js
async function handleSecureInfraRoute({ key, config, json, res, actor }) {
  const op = INFRA_OPERATIONS[key];
  if (!op) throw policyError("not_found", "Unsupported infrastructure operation.", 404);
  requireDiscordCapability(actor, op.capability);         // ADD: capability check
  // ... existing runDune call ...
}
```

Each route handler must pass the actor:

```js
if (path === DISCORD_ADAPTER_ROUTES.SERVERS && req.method === "POST") {
  const body = await readJson(req);
  validateDiscordActor(body.actor);                       // ADD: actor validation
  return handleSecureInfraRoute({ key: "SERVERS", config, json, res, actor: body.actor });
}
```

Same for PORTS and DB.

### Fix 2: /version must use config.repoRoot, not hardcoded path

**Current** (line 198-199):
```js
return json(res, 200, { ok: true, version: readFileSync("/repo/VERSION", "utf8").trim() });
```

**Fixed** — use the existing `readConsoleVersion` from `config.js`:
```js
const version = readConsoleVersion(config.repoRoot);
return json(res, 200, { ok: true, version });
```

### Bot-side verification

The bot already sends actor context with every infra command (`infra:version`, `infra:servers`, etc.) and enforces RBAC before any adapter call. 153/153 bot tests pass post-audit.

Full audit report: https://github.com/yacketrj/dune-awakening-selfhost-discordbot/blob/main/docs/ro-rbac-audit-2026-07-08.md

Red-Blink:
Also this one has conflicts too

Red-Blink:
Please enforce Discord actor/capability checks before running infra commands, and remove the hardcoded `/repo/VERSION`.

## Findings

* `console/api/src/integrations/discord/routes.js:199-210` adds server/ports/db routes that execute `runDune(...)` after bot-token auth only. It reads the request body but ignores `body.actor`, so the advertised capability model is not enforced.

* `routes.js:213-214` hardcodes `/repo/VERSION`; that should resolve from configured repo root.

* Tests do not prove unauthorized actors are rejected.

Red-Blink:
This is closer, but please fix route advertising and add real route tests before merge. 

The new version/servers/ports/db handlers exist, but they are not included in DISCORD_LIVE_ADAPTER_ROUTES, so health reports them as planned instead of live. Update the route list consistently.

Also add tests that actually call the routes and verify missing actor/authz rejection for servers/ports/db, not only constant existence.

PR #82, opened 2026-07-13
fix: orchestrator COPY paths — build context is ./orchestrator, not repo root

yacketrj:
PR #13 changed the orchestrator Dockerfile COPY paths to use `orchestrator/` prefix, but the compose build context is `./orchestrator`, not the repo root. This caused `dune init` to fail with:

```
COPY orchestrator/entrypoint.sh /usr/local/bin/entrypoint.sh
ERROR: failed to calculate checksum: "/orchestrator/entrypoint.sh": not found
```

Fix: restore COPY paths relative to the `./orchestrator` build context (`COPY dune_orchestrator.py`, `COPY entrypoint.sh`).

### DIFF
--- a/orchestrator/Dockerfile
+++ b/orchestrator/Dockerfile
@@ -31,10 +31,10 @@ RUN useradd -m -s /bin/bash dune \
  && chown -R dune:dune /home/dune /srv/dune \
  && usermod -aG docker dune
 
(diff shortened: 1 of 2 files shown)


Red-Blink:
Already fixed

Issue #85, opened 2026-07-17
[Feature] Base Management Page


drkshrk:
# Add "Bases" page under Arrakis Management

## Context

The admin console currently has no way to see player-built bases as a list — the only existing base-related code is `liveMapBases` in `duneDb.js`, which only feeds map markers for the Live Map tab (position only, no piece/placeable counts, no per-base actions).  The goal is to create a dedicated "Bases" tab, listed directly under "Guilds" in the Arrakis Management nav group, showing a sortable table of bases with size/ownership info and per-row actions (Export now, Release Claim as a UI placeholder for later).

This follows the existing `GuildsPanel`/`listGuilds` pattern closely, since Guilds is the most structurally similar existing feature (simple list, no drill-down, per-row actions modeled on `BlueprintsPanel` instead).

## Current Features:
- [x] Add new page to Arrakis Management to display "Bases"
- [x]  Allow exporting a base to a json file stored on your local computer that can then be imported as a Blueprint to any character.
- [x]  Display Generators and Remaining Runtime

## Current Bugs:
- [x] Only displays Fuel Cell and Spice Generators.  The Oil Generators are not displayed.

## Future Iteration / Planned Features:
- [ ] Allow admin to remove ownership of base.  

## Screenshot:

<img width="1994" height="1054" alt="Image" src="https://github.com/user-attachments/assets/1bcb87f5-bc44-423c-b075-7119d46254ad" />

PR #89, opened 2026-07-18
[Feature] Add Base Page

drkshrk:
## Summary

Add a new **Bases** tab under Arrakis Management to view and export player bases and their building pieces.

The new tab displays:
- Base ID, name, owner, and location (map + coordinates)
- Building piece and placeable object counts per base
- Search by base name or owner name
- One-click export to JSON for each base

## Changes

- **Backend** (`console/api/src/duneDb.js`): Added `listBases()` and `exportBase()` functions with proper owner resolution via rank-ordered LATERAL joins and distinct aggregate counts
- **Backend** (`console/api/src/server.js`): Added `/api/bases` and `/api/bases/:id/export` routes with CSRF protection
- **Frontend** (`console/web/src/features/bases/`): New `BasesPanel` component with search and export functionality
- **Tests** (`console/api/test/db.test.js`): Comprehensive test coverage for Bases queries including owner resolution and item counts

## Bug fixes (found in code review)

- Fixed non-deterministic `owner_entity_id` lookup in `exportBase`: now fetches and reuses the owner ID from the base identity query instead of re-deriving it with an unordered subquery
- Fixed owner name resolution in both `listBases` and `exportBase`: replaced `max(character_name)` aggregation with rank-ordered LATERAL subquery to consistently select the actual owner (lowest rank) rather than an arbitrary ranked member

## Testing

All backend tests pass (`node --test test/*.test.js`). Frontend tested in dev server with search, sort, and export functionality.

Refs: #85

### DIFF
--- a/console/api/src/duneDb.js
+++ b/console/api/src/duneDb.js
@@ -2121,9 +2121,14 @@ export async function listBases(db, { q = "", page = 0, pageSize = 50 } = {}) {
   const safePage = intParam(page, "page", 0);
   const offset = safePage * safePageSize;
 
+  // Owner resolution (lowest-rank permission holder) is a per-base correlated LATERAL —
+  // expensive at scale. When searching, the `having` clause needs it to filter on, so it
+  // must run inside `matched` (before pagination) for every candidate base. When not
+  // searching, defer it to the final SELECT so it only runs for the page being displayed.
+  const searching = Boolean(q);
   const values = [];
   let having = "";
-  if (q) {
+  if (searching) {
     values.push(`%${q}%`);
     having = `having coalesce(pa.actor_name, '') ilike $${values.length} or coalesce(owner.character_name, '') ilike $${values.length}`;
   }
@@ -2131,21 +2136,8 @@ export async function listBases(db, { q = "", page = 0, pageSize = 50 } = {}) {
   const limitParamIndex = values.length - 1;
   const offsetParamIndex = values.length;
 
-  try {
-    const result = await db.query(`
-      with matched as (
-        select b.id,
-               a.id as actor_id,
-               pa.actor_name as raw_name,
-               coalesce(pa.actor_name, 'Base ' || b.id::text) as name,
-               coalesce(owner.character_name, '') as owner_name,
-               coalesce(a.map, '') as map,
-               a.transform
-        from dune.buildings b
-        join dune.building_instances bi on bi.building_id = b.id
-        join dune.actor_fgl_entities afe on afe.entity_id = bi.owner_entity_id
-        join dune.actors a on a.id = afe.actor_id
-        left join dune.permission_actor pa on pa.actor_id = a.id
+  const matchedOwnerSelect = searching ? "coalesce(owner.character_name, '') as owner_name,\n               " : "";
+  const matchedOwnerJoin = searching ? `
         left join lateral (
           select ps.character_name
           from dune.permission_actor_rank par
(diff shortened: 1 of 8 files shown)


yacketrj:
Would this be better served on the Players -> Select Player -> Blueprints tab?

Red-Blink:
Thanks for putting this together. The overall Bases page is useful, but I found a couple of areas that should be improved before merging.

The main concern is database load. The page refreshes every 10 seconds, and each refresh runs a fairly large aggregate across buildings, instances, actors, permissions, players, and placeables. On an established server, or with multiple console tabs open, this could generate unnecessary recurring PostgreSQL load.

Please change this to manual refresh or a considerably slower interval, add server-side pagination, and apply filtering before the expensive aggregation where possible. It would also be useful to run `EXPLAIN ANALYZE` against a database containing a realistically large number of structures.

The blueprint export also needs a real round-trip test. It currently reduces rotation to yaw and omits some live-building information, so a complex base containing tilted pieces may not reproduce accurately. The current mocked tests verify the implementation but do not validate the SQL against the actual Dune schema or prove that an exported base imports correctly.

A global Bases page makes sense for server-wide administration because bases can have multiple members. Player-specific navigation can be added separately. The disabled **Release Claim** action should probably be removed until that workflow is implemented.

drkshrk:
> Would this be better served on the Players -> Select Player -> Blueprints tab?

My thinking aligns with Red-Blink's on why I thought it should go in the main structure vs tie into players as it was meant to be a "global" page that captures all Bases across the battlegroup to make viewing and performing global maintenance easier.

I also realized that a lot of similar information is shown via the "Maps" page but that is also showing just "map" specific's.  I think for convenience having a "Bases" tab under players can make sense but was outside the scope of a global base tool. 

One thing i was considering that might make sense but wasn't how I set it up - is that when the server creates a blueprint it is saved server side vs downloading which in that case would tie into the Players->Blueprints page with a list of available blueprints but I didn't want to tackle "server-side" saving as that would have been a little more involved i felt and would have been something I or someone else could try to expand on in a future iteration.  

> The main concern is database load. The page refreshes every 10 seconds, and each refresh runs a fairly large aggregate across buildings, instances, actors, permissions, players, and placeable. On an established server, or with multiple console tabs open, this could generate unnecessary recurring PostgreSQL load.

The database hit for this is a valid point.... i am already like 3 iterations in on revising that as it was originally much slower and I got it a lot faster but didn't take into consideration larger pulls or how often it refreshes....

I think being this is an expensive database call it would make more sense to drastically reduce the refreshing like you mentioned and cache this information unless manually refreshed for something much longer like you said.  I am thinking something like 15 minutes as its not like bases will be flying up and down so quickly.

>The blueprint export also needs a real round-trip test. It currently reduces rotation to yaw and omits some live-building information, so a complex base containing tilted pieces may not reproduce accurately. The current mocked tests verify the implementation but do not validate the SQL against the actual Dune schema or prove that an exported base imports correctly.

I may need some help getting this correct - but when I was working on it the below was what I ran across as I based the export on the blueprint import.  I did test that the blueprint exported and when imported I was able to place it in-game and build it.  I did short-cut and not fully build the structure though but it looked roughly correct. 

### **Known limitation to flag in code comments and to the user:** 
rotation is captured as yaw-only (Z-axis) for both instances and placeable. Any piece/placeable with genuine pitch or roll (tilted geometry) will lose that on export — this mirrors the blueprint format's own inherent limitation for instances (which only ever supported one rotation float), extended here as a deliberate simplification for placeable too, since verifying the exact quaternion→Euler axis convention (order/handedness) without further live testing risked getting it subtly wrong.```

>  It would also be useful to run EXPLAIN ANALYZE against a database containing a realistically large number of structures.

This actually lead down an interesting rabbit hole pointing out some optimization opportunities...  So I appreciate the suggestion to do this as it exposed a bug in the join/combine's
```Massive improvement — 1,078ms → 3.88ms (~278x). ```

drkshrk:
# Bases feature — Current Changes (`feat/Add-Base-Page`)

## 1. Server-side pagination + Shared With fix
[`30304fc`](https://github.com/Red-Blink/dune-awakening-selfhost-docker/commit/30304fc)

- Rewrote `listBases` to page results server-side instead of capping at 500 rows, with search moved back server-side.
- Restored the unfiltered "Total Bases · Total Building Pieces · Total Placeables" summary line.
- Fixed a regression where "Shared With" always showed empty — the pagination CTE split had dropped the actor id, so the join was matching against the wrong id space.
- Added 5 new SQL regression tests targeting the branch's other fragile join-key spots.

## 2. Bases UI polish
[`3552539`](https://github.com/Red-Blink/dune-awakening-selfhost-docker/commit/3552539)

- Added a **Clear** button (inline with Search, before the row-count text).
- Removed the dead, permanently-disabled "Release claim" button.
- Renamed **Export Base → Download Base as Blueprint** throughout:
  - `handleExport` → `handleDownloadBlueprint`
  - `exportBase` → `exportBaseAsBlueprint`
  - `baseExportRoute` → `baseBlueprintDownloadRoute`

## 3. Performance investigation and fix
[`03ce3dc`](https://github.com/Red-Blink/dune-awakening-selfhost-docker/commit/03ce3dc)

- Stress-tested `listBases` with synthetic data (20K–100K bases) via `EXPLAIN ANALYZE` — indexes were fine, but restoring the real production dump (9 bases) exposed two real bugs:
  - A join fan-out between `building_instances` and `placeables` producing 129,301 intermediate rows from just 9 bases (~1,078ms).
  - An owner-resolution `LATERAL` running once per base before pagination applied.
- Fixed both — piece/placeable counts computed independently instead of cross-joined; owner resolution deferred to only the displayed page when there's no active search.
- Verified byte-for-byte identical output on real data; query time **~1,078ms → ~4ms**.

## 4. Deferred (documented, not built)

A further optimization — caching the base list in its own indexed table so pagination scales with page size even at very large base counts — was researched and confirmed feasible, but deliberately not built. Production has 9 bases; this project targets small self-hosted servers, not tens of thousands of bases. Design recorded if needed in future.

PR #102, opened 2026-07-24
Redesign Player Summary with hero header, grouped stats, currency tiles, and Vitals

drkshrk:
## Summary

Overhauls the Player Summary panel in the admin console (`console/web/src/features/players/PlayerSummary.tsx`) and closes out more of the checklist items from #53:

- **Hero header** with character name, online status, map, and guild.
- **Grouped stat tiles** for Level, XP, Skill Points, and Available Intel.
- **New Vitals section** (Health, Hydration, Spice Addiction) with derived max Health from the Vitality passive's Combat-level tiers, plus added max Hydration (100) and max Spice Addiction (10).
- **List-format blocks** Currency (Solari Credit, Solari Coin, Scrip) as their own grouped sections.
- Backend additions: `playerProgression`, `playerIntel`, `playerVitals`, `playerSolarisCoinTotal` in `duneDb.js`, wired through new `/api/players/:id/{progression,intel,vitals,solaris-coin}` routes.

### Robustness fixes bundled in
- Player Summary's data fetch now uses `Promise.allSettled` instead of `Promise.all`, so one failed request (e.g. intel) no longer blanks every other section.
- `playerProgression`/`playerIntel`/`playerVitals` now report their capability as unavailable (instead of defaulting to `0`) when no matching per-player row/component exists, so "no data" is no longer indistinguishable from a genuine zero.
- `playerCapabilities()`'s general capabilities blob now reports the real `progression` support instead of a stale hardcoded `false`.
- Removed a leftover unused `className` prop on `KeyValueGrid` from an earlier, since-superseded nested-box layout.

## Why

Schema research was grounded in a real production DB dump — Dune Awakening's actual survival stats are Health, Hydration, and Spice addiction. The Vitality max-health tier table (Combat level 6/26/56/77/91 → +15/+5/+5/+25/+5, base 150) was confirmed directly by me with access to real game data, since public wiki sources gave conflicting numbers.

## Reviewer notes

- All new read paths gate behind existing `tableExists`/`columnsFor` capability checks and follow the established `resolvePlayerMutationTarget` + parallel-query pattern already used elsewhere in `duneDb.js`.
- No new mutations — everything added here is read-only.
- Live-verified against a restored copy of a real production dump via a throwaway Postgres instance (not included in this PR).

## Test plan

- [x] `node --test test/db.test.js` — 153/153 passing (`console/api`)
- [x] `npx vitest run src/features/players/PlayerSummary.test.tsx` — 23/23 passing (`console/web`)
- [x] `npm run build` (`tsc -b && vite build`) — clean (`console/web`)
- [x] Live-verified in browser against real production data (throwaway DB, never touched the live instance)


drkshrk:
### Rebase Note:
That said, I did find one real (minor) overlap worth flagging:

playerPortalSnapshots inlines its own copy of the Intel query instead of calling our playerIntel(): select coalesce((properties->'TechKnowledgePlayerComponent'->>'m_TechKnowledgePoints')::bigint,0)::text intel from dune.actors where id=$1. Same field we read, written independently, and it has the exact null-masking bug we just fixed in playerIntel (defaults to 0 via coalesce) — just re-introduced in this other, upstream-owned code path. Not something to fix in our PR (it's not our code), but worth knowing it exists if that panel's data ever looks wrong for a player with no TechKnowledge component.
Pre-existing placeholder-text split (not caused by our branch): the portal snapshot's faction/guild fallback comes from addonLeadershipPlayers, which still says "Unassigned"/"Unavailable" — different from our playerProfile's "Neutral"/"—". That inconsistency already existed in the codebase before our branch (I noted it as out-of-scope during the earlier review too); it's just now visible in a second place.

Net: no genuine duplication of what we built, and nothing the rebase mishandled — just one place upstream re-derives the same data point independently, which is a pre-existing upstream pattern, not a merge artifact.

drkshrk:
So I added back the various "Identity's" (Database Index's) -> I dont know if i really like them all there... Maybe would be better to have it expand to show them?  I feel like its useful to be able to easily see them but at the same time they aren't important data unless your trying to do something special. 

I am marking this not-ready for review - i think code-review might have flagged something but i gotta wait till morning to finish it.

image

drkshrk:
Fixed responsive layout issues throughout the Players page.  Everything is now responsive resizing the window - should allow better support for small screens / phones.

Final Layout i settled on....
image

PR #103, opened 2026-07-24
Proposal: add scheduling to the bridge

jeffstokes72:


drkshrk:
### Issues / things worth double-checking before merge

1. **Verify `PAYMENT_SENTINEL_EXPIRY = 999999999` against the real schema.**
   The comment says this must be a "never expires" sentinel for the seller payout row. If `dune.dune_exchange_orders.expiration_time` is a Unix epoch (seconds), `999999999` is **September 2001** — already expired, which would let the game's `dune_exchange_expire_orders` purge the seller's payout and reintroduce exactly the "item eaten, no Solari paid" bug this PR claims to fix. If it's a relative/tick-based field this is fine, but this is the single highest-stakes line in the PR (real player payouts) and isn't verifiable from this repo alone — worth getting written confirmation from the author that it was tested against the live schema, not just ported by inspection from the addon's client code.

### Update on #1 (payment sentinel expiration) — confirmed against the real schema

I restored (a real pg_dump of the live schema) into a throwaway local Postgres and checked this directly. This is **not** a "maybe" — it's a confirmed defect.

### Evidence

**`expiration_time`-style bigint columns in this schema are plain Unix epoch seconds.** No live rows currently exist in `dune_exchange_orders`, but sibling time columns in the same schema do have data:

```
items.acquisition_time              max: 1784254634  → 2026-07-17
player_respawn_locations.last_used_timestamp   max: 1784325659  → 2026-07-17
```

**The real `dune.dune_exchange_purge_completed_orders` function body:**

```sql
CREATE OR REPLACE FUNCTION dune.dune_exchange_purge_completed_orders(in_exchange_id bigint, in_current_time bigint)
RETURNS SETOF exchangeexpiredorder
LANGUAGE plpgsql
AS $function$
...
FOR cur_order IN
    SELECT ...
    FROM dune_exchange_orders ord
    JOIN dune_exchange_fulfilled_orders sord ON (ord.id = sord.order_id)
    WHERE ord.exchange_id = in_exchange_id
      AND ord.expiration_time IS NOT NULL
      AND in_current_time >= ord.expiration_time
    FOR UPDATE
LOOP
    DELETE FROM dune_exchange_orders WHERE id = cur_order.order_id;
    ...
END $function$
```

This join shape (`dune_exchange_orders` ⋈ `dune_exchange_fulfilled_orders`) exactly matches the seller "Take Solari" payment row this PR creates — it's inserted into **both** tables:

```sql
INSERT INTO dune.dune_exchange_orders (..., expiration_time, ..., is_npc_order) VALUES (..., 999999999, ..., FALSE) RETURNING id INTO v_log_order_id;
INSERT INTO dune.dune_exchange_fulfilled_orders (order_id, source_order_id, completion_type, stack_size, original_order_id) VALUES (v_log_order_id, NULL, 4, rec.actual_stack, rec.order_id);
```

### The bug

`in_current_time` is the game server's real current Unix time (~**1.79 billion** right now, per the sibling columns above). `PAYMENT_SENTINEL_EXPIRY = 999999999` is only **~0.999 billion** — i.e. **September 2001**, already in the past relative to any real clock since then.

So `in_current_time >= 999999999` evaluates **true** on essentially every call to `dune_exchange_purge_completed_orders`, meaning the game server's own housekeeping will `DELETE` the payment order almost immediately after the sweep creates it — before the player can collect it.

(I also checked the sibling `dune_exchange_expire_orders` function — it does *not* match this row, since it requires a joined `dune_exchange_sell_orders` row that the payment entry never gets. But `purge_completed_orders` does match, via the `fulfilled_orders` join, so that path isn't a safety net.)

### Fix needed

`PAYMENT_SENTINEL_EXPIRY` needs to be a genuinely far-future Unix timestamp — e.g. `2147483647` (classic 32-bit max, ~2038) or `now() + N years` computed at insert time — for the "never expires" comment/intent to actually hold.

**As written, this will very plausibly reproduce the exact "items eaten, no Solari paid out" bug the PR claims to fix — just delayed by one purge cycle instead of happening immediately.** This should block merge until fixed.

2. The actual expiry logic confirms the sentinel is backwards. Pulled the real function body for dune.dune_exchange_purge_completed_orders(in_exchange_id, in_current_time):

This is exactly the join shape the PR's payment order matches — it's inserted into both dune_exchange_orders and dune_exchange_fulfilled_orders (the seller's "Take Solari" entry). in_current_time is the game server's real current Unix time (~1.79 billion right now, per the sibling columns above). PAYMENT_SENTINEL_EXPIRY = 999999999 is only ~0.999 billion — i.e. September 2001, already in the past relative to any real clock since then.

So in_current_time >= 999999999 is true on essentially every call, meaning the game server's own housekeeping (dune_exchange_purge_completed_orders) will delete the payment order almost immediately after the sweep creates it — before the player can collect it. (I checked the sibling dune_exchange_expire_orders function too; it doesn't match this row since it requires a joined dune_exchange_sell_orders row the payment entry never gets, but purge_completed_orders does match, via the fulfilled_orders join.)

Verdict for the PR comment: this isn't a "please double-check" anymore — it's a confirmed defect. 999999999 needs to be a genuinely far-future Unix timestamp (e.g. 2147483647, or now() + a few years computed at insert time) for the "never expires" comment to actually hold. As written, this sentinel will very plausibly reproduce the exact "items eaten, no Solari paid out" bug the PR claims to fix, just on a delay of one purge cycle instead of immediately.

2. **Minor race on the schedule JSON file.**
   `persistRunCompletion` and `saveBuybackSchedule` are both read-modify-write over the same file with no lock — just atomic rename. The scheduler re-reads before writing completion status (good), but the reverse isn't guarded: an admin's `schedule.set` that reads its `previous` snapshot just before a scheduled run writes its completion status, then writes after, will silently revert `lastRunAt`/`lastRunStatus`/`lastRunDetail` back to stale values. Low impact (display-only fields, narrow window), but worth a one-line comment acknowledging it's accepted, or a small lock.

3. **`scheduler.schedule.set`'s extra permission check reads the raw payload, not the resolved schedule.**
   ```js
   if (payload.enabled) assertInstalledAddonPermission(..., ADDON_SCHEDULER_PERMISSION);
   ```
   If a caller updates other fields (e.g. `intervalMinutes`) on an already-enabled schedule without passing `enabled` explicitly, this check is skipped (the value survives from `previous.enabled` in `normalizeBuybackSchedule`). Not a real security hole since `tick()` still re-verifies `scheduler:server` before every actual run, but a `scheduler:server` revocation isn't reflected in the save-time error until the next tick — a slightly confusing UX gap rather than a vulnerability.

4. **Test coverage gap at the HTTP layer.**
   All 15 new tests exercise `addonJobs.js` directly; none hit `addonSchedulerBridgeAction` in `server.js` (the `id !== EDA_EXCHANGE_BOT_ADDON_ID → 400` branch, the permission/rate-limit wiring per action, or the `scheduler.run` → `addonJobScheduler.runNow` plumbing). Given `server.js` is one of the two largest files in the repo, some route-level coverage would reduce risk of regressions there.

5. **Minor validation inconsistency**: `intervalMinutes` silently clamps to `[10, 1440]` while `priceMultiplier`/`buybackPercent`/`maxBuys` throw on out-of-range. Intentional per the docs table, but worth flagging since one bad field silently coerces while another errors — slightly surprising for addon authors integrating against this API.

### Nitpicks
- `purchased`/`totalUnits`/`totalSolari` are parsed with plain `Number()` from DB rows, while the file otherwise treats exchange-related BIGINTs with explicit overflow-safety (string + `BigInt`). Practically bounded by `maxBuys` (≤5000) so unlikely to matter, but slightly inconsistent with the file's own stated principle in the header comment.
- Placement of `addonJobs.js` at `src/` top-level (rather than `src/services/`) matches its sibling `addonItemGrants.js`, so that's fine/consistent.

### Overall
Solid, unusually well-tested addition that follows this repo's existing addon-bridge conventions (permission-gated writes, `redact()`+`audit()` on every branch, backup-before-write). 1 Major issue and several minor to consider.

jeffstokes72:
Thanks for the review, will get it cookin in the morning

jeffstokes72:
Cursor is telling me that time isn't unix epoch time, the original code has a long writeup on why thats the case and how the time is inferred. Gonna dig that up tomorrow but thought I'd resubmit after running it through 2 cloud agents with different frontier models and both telling me same.

PR #109, opened 2026-07-26
ci: add Compose project-name portability test, wire up dead command-auth-token test

yacketrj:
## Summary

Two independent CI gaps, both found during a review of why recent PRs
required post-merge reverts (command-auth-token in July, and PR #108's
Compose `name:` pin reverted 11 minutes after merge by 443152d).

1. **`runtime/tests/test-compose-project-name-portability.sh` (new)**

   Reproduces the exact PR #108 bug class mechanically instead of
   relying on someone remembering to ask "what if this operator's
   directory has a different name?" It creates a real `git worktree`
   under a deliberately different basename and asserts, via
   `docker compose config`, that the resolved project name (and
   therefore every named volume) still tracks the directory basename,
   not a hardcoded string. It also statically checks that no compose
   file with top-level named volumes hardcodes a `name:` field --
   scoped to files with real persistent volumes at stake, so it does
   **not** flag `docker-compose.web.yml`/`docker-compose.public-probe.yml`,
   which intentionally pin `name:` but declare zero named volumes
   (bind mounts + explicit `container_name:` only), so they carry
   none of the orphaning risk this test exists to catch.

   **What this does not cover**: only Compose's own project-name
   resolution logic; it does not run a full install or exercise
   `self-update.sh`'s upgrade path end-to-end against a differently
   named directory. That remains a manual step until a disposable
   upgrade-test environment exists.

2. **Wire `runtime/tests/test-command-auth-token.sh` into `ci.yml`**

   This test already existed but was never referenced by any CI
   workflow -- a regression guard that never actually ran. Added to
   the `runtime-script-unit` job alongside the new test above.

## Test plan

- Both scripts run standalone (`bash runtime/tests/*.sh`) and via
  `shellcheck -S warning` (this repo's CI severity) with zero findings.
- Full `.github/workflows/ci.yml` re-validated as parseable YAML.
- Real CI run on the fork, triggered via `workflow_dispatch` (branch
  name doesn't match the `push` trigger's filter): **all 10 jobs
  passed** (`web-tests`, `api-tests`, `metrics-unit`,
  `runtime-script-unit`, `container-lifecycle`, `security-checks`,
  `api-dependency-audit`, `release-validation` on both ubuntu-22.04
  and ubuntu-24.04, `release-gate`) --
  https://github.com/yacketrj/dune-awakening-selfhost-docker/actions/runs/30185252541
- Live stack on the authoring host was not touched or restarted at
  any point.

## What's intentionally not in this PR

A full upgrade-path test against a real disposable staging deployment
-- no such environment exists yet on the authoring host; tracked as a
known gap, not silently assumed to be covered by this PR.


Red-Blink:
Thanks, the intended CI protections are useful, especially the Compose project-name regression test. However, the PR branch currently contains 39 commits, 32 changed files, and more than 6,000 unrelated added lines from previous work.

Please rebuild or rebase the branch onto the current `Red-Blink/main` so that it contains only commit `19597586de49`.

Please also:
* Dynamically discover all `docker-compose*.yml` and `compose*.yml` files instead of hardcoding four filenames.
* Strengthen `test-command-auth-token.sh` so it functionally verifies that `DUNE_COMMAND_AUTH_TOKEN` overrides the fallback, rather than only checking that the relevant strings exist.

I tested the intended commit by itself against the current base, and the runtime test suite passed. Once the branch history and these test gaps are corrected, it should be suitable for review.

yacketrj:
Closing -- superseded by upstream's own work. While this PR was open, upstream independently built a more complete version of the same fix in commits `9d09ec7` ("fix(compose): preserve existing project volumes") and `7d10a41` ("test: enforce release quality regressions"), released as part of v1.3.66. That work includes a real `runtime/scripts/compose-project.sh` discovery/persistence library, an evolved version of this PR's own `test-compose-project-name-portability.sh`, a complementary `test-compose-project-resolution.sh`, and the exact CI wiring (including `test-command-auth-token.sh`) this PR proposed.

Verified directly: after syncing our fork with `upstream/main` (v1.3.66) and rebasing this PR's branch on top, the resulting diff against `main` was empty -- everything this PR contributed is already present. No further action needed on this PR.

PR #110, opened 2026-07-26
fix(bases): Add missing turbines, correct generator fuel runtime and clarify depletion status

drkshrk:
## Summary
- Corrected generator fuel-burn rates (Oil, Spice, both wind-turbine lubricant grades) and fixed runtime reporting so a generator that verifiably holds fuel no longer reports 0 — root causes were an idle-state string sentinel matched against literally, wrong burn-rate constants, and unrecognized placeable variants silently defaulting to the Fuel type.
- Added the two wind turbine types as their own generator entries, and a base with an unfuelled generator now reads "out of fuel" instead of "0m left".
- Verified the previously-unmeasurable Spice fuel cell burn rate via live measurement once a spice generator existed on the server.
- Reworked the Bases page's Generators column to show "next depletion in X" (the soonest a still-fuelled generator runs dry) and how many generators are empty, rather than one minimum getting dragged to 0 by any single empty generator. Added tooltips for truncated text throughout the shared DataTable component.

## Test plan
- [x] `cd console/api && node --test test/db.test.js` — 175/175 passing
- [x] `cd console/web && npm run build` — clean
- [x] `cd console/web && npx vitest run` — 75/75 passing
- [x] Rebuild the console image and confirm per-base generator counts match pre-change numbers in production


drkshrk:
image

Issue #112, opened 2026-07-28
Ability to "refill" Power generators.


drkshrk:
### Status:
So I have an initial implementation of this.  As expected it does require you "restarting" the server.   Still needs some work(UI/Feedback from it) but the feature is there and it seems to work correctly.
https://github.com/drkshrk/dune-awakening-selfhost-docker/tree/feat/GenRefill

- Verified Clicking Generator Refill makes DB edits and fills all generators per the "Max" settings that I specified.  After restarting the map generators that were empty are now full.
- Built and tested systems to support.  Need to do final review of UI and make sure it looks good before submitting PR.

### Pre-Req to adding:
- [x] 1. Need to be able to manually restart each Sietch Individually.  I currently see a "Force Despawn" for Deep Dessert and Social Hubs, but nothing on Sietch's.
- [x] 2. Build a Queue system for "holding" generator refills until a map restarts.  This is done to prevent any chance of a running server overwriting the database edit.  This further supports the "Automatic" refill option.
- [x]  3. Add "Auto re-fill button and wire into queue system.

# How it Works:
## Refill caps

| Device (`GENERATOR_TYPES` key) | Fuel template | Fuel name | Stack size | Max stacks | Total cap |
|---|---|---|---|---|---|
| `fuel` | `Oil` | Fuel Cell | 499 | 1 | 499 |
| `spice` | `SpicedFuelCell` | Spice-infused Fuel Cell | 499 | 1 | 499 |
| `windTurbineOmni` | `WindTurbineLubricant1` | Low-grade Lubricant | 100 | 5 | 499 |
| `windTurbineDirectional` | `WindTurbineLubricant2` | Industrial-grade Lubricant | 100 | 5 | 499 |

Turbines cap at 499 across 5 stacks → 4×100 + 1×99.

**Configurable, not hardcoded.** Defaults sit alongside the existing type definitions; an optional `runtime/data/generator-refill-caps.json` overrides any field by type key, merged at load. This mirrors how `runtime/data/admin-items.json` is already consumed by `console/api/src/adminCatalog.js`, so operators can retune caps without a rebuild.

## API route

**`console/api/src/server.js`**, beside the existing base export route at `:429`:

```js
if (path.match(/^\/api\/bases\/[^/]+\/refill-generators$/) && req.method === "POST") {
  return baseRefillGeneratorsRoute(req, res, path);
}
```

```js
async function baseRefillGeneratorsRoute(req, res, path) {
[...]

PR #116, opened 2026-07-31
feat(bases): generator refill queue + per-base auto-refill toggle

drkshrk:
## Summary

- **Manual generator refill** (`POST /api/bases/:id/refill-generators`): tops every power device on a base (fuel/spice generators, omni and directional wind turbines) to a configurable cap in one transaction. Caps are overridable via `runtime/data/generator-refill-caps.json`.
- **Refill queue**: a refill is queued instead of applied immediately when the target map isn't safely writable right now (a live game server can overwrite the base back to Postgres on its own timer). Queued entries flush automatically once their map/partition is confirmed down, are visible via a pending-refills banner across Bases/Maps/Server panels, and can be canceled. Adds a "restart via despawn+respawn" path (`POST /api/maps/respawn`) for maps with no managed service.
- **Per-base Auto-Refill**: an opt-in per-base scheduler (`console/api/src/services/autoRefill.js`) that periodically checks generator fuel levels and queues a refill automatically when any generator drops below a configurable threshold. Surfaces enrollment/rule explanation via a shared `InfoTooltip` popover (matching the Maps panel's memory-toggle pattern), and detects a base whose refills keep failing to raise fuel (stalled after 3 consecutive attempts), shown via a red row icon and a page-top banner.
- **UI polish** (several rounds, iterated against live production data): toggle sizing, moving the "2× Uptime Event" badge inline with the toggle, tooltip open-direction fix, a queued-refill actions-column overflow fix, and a mobile tooltip-inset fix.
- **Docs**: `docs/API-REFERENCE.md` (route reference for the new endpoints), `docs/generator-refill-caps.md`, `docs/generator-fuel-burn-rates.md`.

Addresses #112 (manual refill) and #85 (auto-refill toggle + polish).

## Test plan

- Backend: `console/api/test/db.test.js` (216 tests) and `console/api/test/autoRefill.test.js` (22 tests) — mocked-`db` unit tests covering the refill transaction, queue store/cancel/flush, write-safety dwell logic, and the auto-refill scanner/stall detection. Run directly (no live Postgres needed for these files).
- Frontend: `npx vitest run` — 107/107 passing; `npm run build` (`tsc -b` + `vite build`) clean.
- Live-verified on the production host across multiple deploys: each `sync-branch.sh -origin` dry run matched the expected changed-file list before applying, and after each `-live` apply all 6 game server containers kept their pre-deploy uptime (only the console/orchestrator management containers recreate) with the console responding 200 post-rebuild. Auto-refill and stalled-state detection were verified against existing production data at known fuel levels rather than writing test data.
- A 4-agent parallel pre-commit review (schema/data, frontend, security, hygiene) was run against `upstream/main` before opening this PR; its actionable findings (actions-column overflow, a CSS specificity bug, a missing log-redaction call, and 5 undocumented API routes) are fixed in the history below.

## Notes for reviewers

- `console/api/src/duneDb.js` and `console/api/src/services/autoRefill.js` are the two files to focus on for the write-safety logic (a queued refill only applies once a map is confirmed down via `pg_stat_activity`, never straight into a possibly-live base).
- Draft PR — flagging for review before marking ready.


Red-Blink:
Thanks for the work on this. I reviewed head `ab2dd6a`; all CI checks and local tests pass, but these issues need fixing before merge:

1. **Concurrent empty-inventory refills**

   `refillBaseGenerators()` only locks existing fuel rows. An empty inventory has nothing to lock, so concurrent manual requests can both calculate the full deficit and insert duplicate fuel.

   Lock the inventory/placeable row or use a transaction-level advisory lock before reading fuel, and add a real concurrency test for an empty generator.

2. **Map writes must flush between stop and start**

   `flushPendingMapWrites()` runs after `runDune()` finishes. For `restartService` and `sietchesRestart`, the map has already been stopped and restarted by then, so the safe write window can be missed.

   Please make these explicit stop → flush → start operations instead of relying on the five-second poller.

3. **Auto-refill stall timing**

   Auto-refill sets `stalledAt` as soon as the third refill is queued, before that refill is applied.

   Only mark it stalled on a later scan when the third queue has completed/disappeared and fuel remains low.

4. **Validate refill timing settings**

   Validate and clamp:

   * `ADMIN_REFILL_RETRY_DELAY_MS`
   * `ADMIN_REFILL_DOWN_DWELL_MS`

   Raw `Number(...)` permits invalid or negative values that can break retry and dwell safety.

5. **Correct the API documentation**

   The API reference says database writes create automatic backups, but `directDbMutation()` returns `backupCreated: false`.

   Please correct the documentation.

Please add tests covering concurrent empty-inventory refills and guaranteed flushing between map stop/start. After those are addressed, I’ll review the updated head.

drkshrk:
Got it - I will address these this evening.  Had to go into the office today and my VPN is apparently not happy.  Hopefully I will have updated for you to review tomorrow.

PR #133, opened 2026-08-08
Add a read-only Base Inventory tab

drkshrk:
Adds a read-only Inventory tab to the expanded base row, showing what a base
holds — storage containers plus refining, crafting and machine inventories —
rolled up either by item template or by container. Backed by a new
`GET /api/bases/{baseId}/inventory`.

Classification is an explicit `building_type` allowlist rather than
`inventory_type`: Recycler and Repair Station are type 3, the same as the oil
generators the Power tab owns. All 24 types were verified against the shipped
server paks and a live database; unrecognised placeables are omitted rather
than guessed at.

Read-only by design — item writes have no live-sync path: no `pg_notify`
channel carries them, there are no triggers on `dune.items`/`dune.inventories`,
and the RMQ command bus addresses items by template name while every id here is
a row id.

Also in this branch:

- **Map column** now identifies *which* instance of a map a base is on. It
  showed `dune.actors.map`, which is identical for every instance — a server
  running two Deep Desert instances showed both as "DeepDesert". Resolved
  through `world_partition` and labelled with the operator-chosen instance
  name, falling back to the partition number.
- **Sietch partition ids** — `parseSietchRows` falls back to the dimension
  index when the `--ids` output is short or empty, and those indices were
  pooled across all maps, so one map's dimension index could label another
  map's base. The same fallback reached every sietch write: a rename, password
  change or restart could land on a different partition than the one being
  edited. Such rows are no longer valid write targets, and the save is refused
  with a message rather than skipped — a bulk Save that skipped them either did
  nothing at all or applied an unrelated active-count change while silently
  discarding the edited fields. Reads are unaffected; the sietch list still
  renders in full.
- **`itemImagePath`** ran an uncached `existsSync` per item id, so a catalog
  response repeated up to 10,000 synchronous stats per call.
- **`bases-generator-*` card classes** renamed to `bases-card-*` now that three
  tabs share them. No visual change.

Docs in `docs/console/base-inventory.md`, with `API-REFERENCE.md` and the docs
index updated.

271 frontend tests pass; the backend suite is green on Linux apart from 5
pre-existing failures that require a live database, unchanged from baseline.

### DIFF
--- a/console/api/src/adminCatalog.js
+++ b/console/api/src/adminCatalog.js
@@ -99,8 +99,25 @@ const itemImagePathCache = new Map();
 // checkout at /repo and points DUNE_DOCKER_DIR at it, so the files are live on
 // disk. What makes the stale window harmless is that updating the checkout
 // restarts the console container, not that the files cannot change.
+// An id becomes both a filesystem path and a URL below, so it may not carry
+// path structure. normalizeItem's own id regex does not cover this -- it admits
+// "." and "/", so "../../secret" satisfies it -- and baseInventory passes a raw
+// dune.items.template_id with no validation at all. Without this, resolve()
+// normalises the traversal away and existsSync probes outside the public
+// directory, while the returned string reaches an <img src>.
+//
+// Rejected ids resolve to the unavailable image rather than throwing, matching
+// what a missing file already does. Checked before the cache so a malformed id
+// cannot take a slot in it either.
+function isSafeItemImageId(id) {
+  const value = String(id ?? "");
+  if (!value || value.length > 240) return false;
+  if (/[\\/\0]/.test(value)) return false;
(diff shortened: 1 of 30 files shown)


Red-Blink:
Found one failure-path issue in the Sietch write protection. `survivalSietchActions()` silently skips rows whose partition IDs are unreadable. If only the name/password is dirty, Save does nothing; if the active count is also dirty, that change can succeed while the edited Sietch fields are silently discarded.

Please make bulk Save surface `SIETCH_PARTITION_IDS_UNREADABLE` whenever a dirty untrusted row exists, matching the individual Save/Restart behavior, and add regression coverage for both cases.

drkshrk:
ok - should be ready for re-review... rabbit trail of things i came across

PR #153, opened 2026-08-14
feat(secrets): age-based envelope encryption for Postgres, Funcom token, RMQ secret, FLS API key, and server-login secrets

yacketrj:
## Summary

Adds a general-purpose, opt-in age-based secrets manager
(`runtime/scripts/lib/secrets.sh` + `secrets_aead.py`) and wires it up for six secrets that
were previously read as plain flat files, scattered across more than a dozen scripts:

- The Postgres superuser password (previously `-e POSTGRES_PASSWORD=postgres`, visible in
  full via `docker inspect --format '{{.Config.Env}}'` to anyone with Docker socket access —
  GHSA-fc89-h24v-6j3x)
- The Funcom service token
- The RabbitMQ HTTP token-auth secret
- The FLS API key
- The server-login password secret and username-server-login secret

**Strictly opt-in, per secret.** An operator who sets neither `DUNE_KEK_FILE` nor
`DUNE_AGE_IDENTITY_FILE` sees zero behavior change for any of the six — every one of them
continues to be read from its existing plain flat file, exactly as before. You do not have
to migrate all six at once; each is opted into independently.

**This PR treats the upgrade path — not just the fresh-install path — as the primary design
constraint, and documents that verification in more detail than is probably typical for a
PR this size.** I want to be upfront about why: while building this, I found and fixed a real
bug where enabling the feature against an *existing* Postgres installation would silently
break authentication (see below). That experience changed how I approached everything after
it — every one of the other five secrets was re-verified against a live, running,
already-populated deployment specifically to rule out the same class of failure, not just
reasoned about by analogy. Full evidence is in the new documentation file
(`docs/runtime/AGE-SECRETS-MANAGEMENT.md`, "Upgrade-Path Verification Evidence" section) —
I'd genuinely appreciate scrutiny on that section in particular, since it's the part most
likely to matter to someone else's real, live server.

## What's in this PR

**New library files:**
- `runtime/scripts/lib/secrets_aead.py` — AES-256-GCM primitive via Python's `cryptography`
  package. Required because `openssl enc` cannot do AEAD ciphers at all (confirmed:
  `enc: AEAD ciphers not supported`, permanent upstream policy, not a version gap) and Bash
  has no native AES-GCM. Secret material is passed exclusively via stdin, never argv —
  regression-tested via real subprocess `/proc/<pid>/cmdline` inspection.
- `runtime/scripts/lib/secrets.sh` — the shared orchestration seam, meant to be sourced like
  `host-paths.sh`/`runtime-env.sh`. KEK decryption via `age` (cached in-process only, never
  written to disk), atomic write-then-fsync-then-rename for secret writes, transparent
  fallback to `runtime/secrets/*.txt` when unconfigured. The generic
  `dune_secrets_read_secret <name> <legacy-path>` function is what every one of the five
  non-Postgres secrets goes through; Postgres has its own dedicated
  `dune_secrets_sync_postgres_password()` (see below for why).
- `runtime/scripts/lib/test_secrets_aead.py`, `runtime/tests/test-secrets-aead-cross-language.sh`,
  `runtime/tests/test-secrets-lib.sh` — unit/integration coverage for the crypto primitive
  and the shared library, including a real cross-language (Python/Node) ciphertext round-trip
  proof.

**New resolver functions in `runtime/scripts/runtime-env.sh`:**
- `resolve_funcom_token`, `resolve_rmq_http_token_auth_secret`, `resolve_fls_apikey`,
  `resolve_server_login_password_secret`, `resolve_username_server_login_secret` — one shared
  resolver per secret, replacing the inline `TOKEN_FILE="runtime/secrets/....txt"` +
  `tr -d '\r\n' < "$TOKEN_FILE"` pattern that was duplicated near-identically across 7+
  scripts. Two of these (`resolve_rmq_http_token_auth_secret`, `resolve_fls_apikey`) preserve
  the existing auto-generate-if-missing behavior (`openssl rand -hex`) that 3-5 of the
  consuming scripts already had — but only generates a *new* flat-file value when the
  encrypted form isn't already authoritative for that secret, so a migrated-then-flat-file-
  removed operator never has it silently regenerated out from under the encrypted value.
- `funcom_token_host_id()` (parses a JWT to extract `HostId`) refactored to accept the token
  **value** directly, not just a file path — it previously only worked by reading a file,
  which would have forced writing a decrypted token to a temp file just to parse it once
  migrated. Piped via stdin, never argv or an environment variable (see "Two bugs caught
  before they shipped" below for why this distinction mattered in practice, not just in
  theory).

**Migrated call sites:** `spawn-server.sh`, `start-director.sh`, `start-server-gateway.sh`,
`start-server-overmap.sh`, `start-server-survival-1.sh`, `start-text-router.sh`,
`admin-tools.sh` (existence check only), `db.sh` (a battlegroup-mismatch diagnostic),
`doctor.sh` (health-check display), `battlegroup-identity.sh`.

**Modified:**
- `runtime/scripts/start-postgres.sh` — moves the superuser credential to
  `POSTGRES_PASSWORD_FILE`; adds `dune_secrets_sync_postgres_password()` (see below); tightens
  `01-create-dune-user.sql` to 0600 (previously unset, defaulting to umask's usual 644 —
  world-readable, containing the `dune` role's password in plaintext SQL).
- `.pre-commit-config.yaml` — removes `pass_filenames: false`/`always_run: true` from the
  `semgrep` hook, which caused every commit to trigger a full-repo scan regardless of what
  was actually staged.
- `tests/security-pr-checks.sh` — adds `-x` to `shellcheck`'s invocation and adds this
  feature's new test files to it.
- **New documentation**: `docs/runtime/AGE-SECRETS-MANAGEMENT.md` — full operator-facing
  install/usage/break-glass guide, including the detailed upgrade-path verification evidence
  mentioned above.

## The real upgrade-path bug (Postgres-specific) — the reason this PR is structured the way it is

Postgres's own `docker-entrypoint.sh` only applies `POSTGRES_PASSWORD_FILE` during
**first-time** `initdb`. Enabling this feature on a host with an **existing,
already-initialized** Postgres data directory — i.e. any real, existing deployment, not a
brand-new install — silently generated a new password and wrote it to the encrypted store,
but Postgres never actually adopted it, because the data directory already existed. The live
database's real password stayed whatever it was at original init time, while the encrypted
store held a value that didn't match reality, with no error at the point the bad value was
written.

Reproducing this correctly required discovering a real verification pitfall: this image's
`pg_hba.conf` grants unconditional `trust` for local (`127.0.0.1`/`::1`/`local`) connections.
Verifying via `docker exec`/localhost gives a false "success" for *any* password, correct or
not — the actual reproduction required a real client on the container's Docker network.

**Fixed** via `dune_secrets_sync_postgres_password()`: after every container startup where
the age backend is configured — not just the first one — the live password is explicitly
forced to match the encrypted store's value via `ALTER USER ... WITH PASSWORD ...` over the
same trusted local connection. If this sync fails, the script exits non-zero with a loud,
actionable error rather than silently reporting success with a diverged secret. Regression-
tested against a real, disposable Postgres container simulating a genuine pre-existing
install (`runtime/tests/test-postgres-secrets-upgrade-path.sh`).

That same test file also carries a deterministic regression test (`strace -f -e trace=execve`
capturing real process argv at the moment of `execve(2)`) guarding against
`dune_secrets_sync_postgres_password` ever passing the password via
`psql -c "ALTER USER ... '...'"` instead of piping it over stdin — that form would put the
plaintext password directly into a process's own argv, the exact GHSA-fc89-h24v-6j3x exposure
class this feature exists to eliminate, reintroduced in a different location. I found and
fixed this in my own first draft before it ever shipped, by independently reproducing the
exposure live via `/proc/<pid>/cmdline` inspection.

**Why doesn't this same bug class apply to the other five secrets?** This is explained in
detail in the new documentation, but the short version: the Postgres bug exists because
Postgres's own startup code makes an internal "apply this password, or don't" decision based
on data-directory state this project's scripts have no visibility into. The other five
secrets have no equivalent internal state — they're plain values delivered fresh via
`docker run -e` on every container start, re-read from scratch every time, with no analogous
"already initialized, ignore this" logic anywhere in the consuming game-server binary. There
is structurally no way for the encrypted store and a live running process to silently diverge
for these five the way there was for Postgres. I did not take this argument on faith, though
— see the verification evidence below.

## Two bugs caught during development of the five-secret migration, before either reached a live host

1. **A Python heredoc silently discarding a piped value.** `funcom_token_host_id()`'s internal
   helper originally combined `printf '%s' "$token" | python3 - <<'PY' ... PY` — confirmed
   directly that a heredoc takes over `python3`'s stdin and silently discards the pipe
   (`printf x | python3 - <<'PY' ...` reads the heredoc body, never the pipe). Fixed by
   passing the script via `-c` instead, leaving stdin free for the actual token.
2. **An environment-variable-shaped exposure, caught in review before it shipped.** An earlier
   draft of the same refactor considered passing the resolved token to that Python helper via
   an environment variable. Rejected: `/proc/<pid>/environ` is visible to any other process
   running as the same user, for that process's entire lifetime — a materially real, if
   different, exposure to the argv-based one (GHSA-fc89-h24v-6j3x) this whole feature exists
   to close. Fixed to pipe via stdin instead; confirmed live via `strace -f -e trace=execve`
   that the shipped version introduces zero such exposure (the traced `execve` for the Python
   subprocess shows only the fixed, non-secret script text — the token itself never appears
   in any captured `execve` call).
3. **A subtle argument-handling bug that could mask a real failure as a stale success.**
   `funcom_token_host_id()`'s original `${1:-default-path}` treats an *explicitly passed empty
   string* (e.g. the result of a failed `resolve_funcom_token` call) identically to *no
   argument at all*, silently falling back to re-reading the default flat file — meaning a
   genuine resolver failure could be masked by successfully reading a different, stale value
   instead of the failure propagating as an error. Fixed to distinguish "not passed" from
   "passed as empty string." The regression test for this
   (`runtime/tests/test-runtime-env-secrets-resolvers.sh`) required a real, synthetic,
   parseable JWT planted at the exact default relative path to actually prove the fix closes
   the gap — a first attempt at this test, without that fixture, passed even with the bug
   still present (for the wrong reason: a missing file, not the intended check), which was
   caught and corrected during test development rather than assumed sufficient.

## Testing

- `runtime/scripts/lib/test_secrets_aead.py`: 18/18 pass
- `runtime/tests/test-secrets-aead-cross-language.sh`, `test-secrets-lib.sh`,
  `test-postgres-secrets-upgrade-path.sh`, `test-runtime-env-secrets-resolvers.sh`: all pass
- `shellcheck -x`: clean on every new file and on the shared library/test files added to
  `tests/security-pr-checks.sh`'s invocation. (The 12 modified production game-server scripts
  carry pre-existing, unrelated shellcheck findings that predate this change — confirmed via
  `git diff`/history comparison — and were deliberately **not** added to that CI invocation
  in this PR, to avoid either silencing 15+ unrelated findings inline or fixing unrelated code
  as incidental scope creep. `bash -n` syntax-checked and manually reviewed instead.)
- **Live-verified against this project's own real, production self-hosted server** (population
  confirmed at 0/60 throughout, explicit operator authorization to proceed regardless of
  population for this maintenance-mode verification work): for every one of the five
  non-Postgres secrets, captured the actual delivered container environment variables,
  restarted the relevant script(s) with the migrated code and the age backend deliberately
  left **unconfigured** (the real state every existing operator is in on upgrade), and
  confirmed byte-for-byte identical secret delivery before and after. Full table, secret by
  secret and container by container, in `docs/runtime/AGE-SECRETS-MANAGEMENT.md`'s
  "Upgrade-Path Verification Evidence" section.
- **A real, unrelated incident was found during this live verification** and is documented in
  full in that same section, including why it was investigated and ruled out as unrelated
  (exact log timestamps showing the problem predated this session by 3+ hours; no correlated
  Postgres-side event; clean recovery using the exact migrated code under test) rather than
  silently omitted. I'd rather over-disclose here than have this look like something was
  swept under the rug.

## Explicitly out of scope

- The `dune` role's own application password, as consumed via `-DatabasePassword=` argv by
  the world-server binary, is a separate, harder problem — env-var delivery was tested twice
  against that binary and failed both times.
- `admin-web-password.txt`, `admin-web-session-secret.txt`, and `public-directory.json` are
  Node.js/console-consumed (`config.js`), not shell-consumed — out of scope for this Bash
  library.
- `discord-adapter-token.txt` has zero live references anywhere in the current codebase —
  nothing to migrate.
- `command-auth-token.txt` is declared but unused in `admin-tools.sh` (dead code) — not
  migrated. This specific secret has its own documented history of real production incidents
  from token-generation/rotation changes; touching it deserves its own explicit, separate
  decision, not incidental inclusion here.
- A dedicated `dune secrets setup` CLI command to automate the manual opt-in steps is a
  natural follow-on but not part of this change.
- Disabling the Postgres migration after enabling it does not currently revert the live
  database's password back to the hardcoded default (documented explicitly, with the
  workaround, in the new doc's Break-Glass Recovery section). This limitation does **not**
  apply to the other five secrets — unsetting the two env vars immediately and correctly
  reverts all of them to their flat files, since none of them are held inside a long-running
  process the way the Postgres password is.


Red-Blink:
Thank you for the work and detailed testing behind this. Protecting stored credentials is a worthwhile goal, but this PR changes too many security-critical systems at once for me to merge safely in its current form.

It affects PostgreSQL authentication, Funcom tokens, RabbitMQ, FLS, server-login credentials, startup scripts, diagnostics, backups, and upgrade behavior. A regression in any of these areas could prevent an existing battlegroup from starting or make its credentials difficult to recover.

Please keep this as a draft and split it into smaller, independently reviewable changes. I’d suggest starting with the shared secrets library and one lower-risk credential, then handling PostgreSQL separately because its upgrade and rollback behavior is substantially more sensitive.

Each stage should cover:

- Fresh installations
- Existing installations upgrading from plaintext secrets
- Disabling the feature and returning to the previous configuration
- Missing, incorrect, or lost encryption keys
- Backup and disaster recovery
- Partial migration failures
- Restarts and automatic updates
- Confirmation that secrets do not appear in arguments, logs, temporary files, or unintended environment output

The direction is promising, but this needs a smaller scope and careful security review before it can be considered ready to merge.

yacketrj:
Reverting back to draft per your review comment above -- this should stay draft until it's properly superseded by the smaller, independently reviewable split you requested. Stage 1 of that split (the shared secrets library, with zero call-sites/zero behavior change) is now up as #160. Will follow up here once the split is far enough along to close this PR out.

Red-Blink:
This original combined secrets PR has now been superseded by the staged split beginning with #160, and it conflicts with current main in `runtime/scripts/runtime-env.sh`.

Please close this PR rather than continuing to update it. Keeping the superseded implementation open makes the security review and intended merge path unclear.

yacketrj:
Closing as requested -- superseded by the staged split (#160 and onward). Thanks for the guidance on how to split this safely.

PR #154, opened 2026-08-14
fix(installer): preserve .env mode/ownership on Alpine without GNU-only chmod flag

drkshrk:
## Summary
- `dune_persist_compose_project_name()` used `chmod --reference=` / `chown --reference=` to preserve `.env`'s mode/ownership before the atomic replace. `--reference=` is GNU-coreutils-only; BusyBox `chmod`/`chown` on Alpine don't support it and abort with a usage error.
- `choose_web_port()` already creates `.env` earlier in the same install run, so this hit on every fresh Alpine install (not just upgrades), crashing the installer right after "Starting the Web UI." — before Docker Compose ever ran.
- Replaced with a portable `stat -c '%a'` / `stat -c '%u:%g'` capture plus plain `chmod`/`chown`, which works on both GNU coreutils and BusyBox.

## Test plan
- [x] Reproduced the exact crash on a clean Alpine 3.22 container using real BusyBox `chmod`/`chown`
- [x] Applied the fix and reran the same install end-to-end — completes fully, console starts, admin password issued
- [x] Confirmed `.env` mode/ownership are still correctly preserved after the fix
- [x] `runtime/tests/test-compose-project-resolution.sh` passes with the new implementation (updated its stale assertion + added a regression check for `--reference=`)
- [x] `shellcheck` / `dash -n` clean on both changed files

### DIFF
--- a/runtime/scripts/compose-project.sh
+++ b/runtime/scripts/compose-project.sh
@@ -203,8 +203,16 @@ dune_persist_compose_project_name() {
     # Scheduled systemd helpers can run as root. Preserve both mode and
     # ownership before the atomic replacement so a root-created temporary file
     # cannot turn the operator's .env into root-owned state.
-    if ! chmod --reference="$dune_compose_env_file" "$dune_compose_tmp_file" \
-      || ! chown --reference="$dune_compose_env_file" "$dune_compose_tmp_file"; then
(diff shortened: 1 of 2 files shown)


drkshrk:
Alpine install was failing due to this -> this got it passed where it was failing.  I havent verified full deployment though as of yet.

PR #155, opened 2026-08-14
fix: keep sidebar nav reachable on tall pages

drkshrk:
## Summary
- The sidebar had no height cap and `overflow: visible`, so `position: sticky` couldn't keep both its top and bottom pinned once the nav list grew taller than the viewport — lower nav items only became reachable once the main content was scrolled almost to the bottom.
- Locks the sidebar to one viewport height and drives its scroll position off the page's own scroll progress, so nav items reveal themselves gradually from the first scroll instead of sitting frozen until near the end, with no separate scrollbar on the sidebar.

## Test plan
- [x] Verified live against a real restored test dataset (Docker Postgres 17.4 + console build) that the sidebar stays pinned to the page scroll regardless of main content height
- [x] Verified sidebar `scrollTop` tracks page scroll progress linearly (0/25/50/75/100%)
- [x] Verified no visual regression on a viewport taller than the nav content
- [x] Verified the `max-width: 900px` mobile top-bar layout is unaffected
- [x] Deployed and confirmed live on a real self-hosted instance

### DIFF
--- a/console/web/src/App.tsx
+++ b/console/web/src/App.tsx
@@ -312,6 +312,7 @@ export function App() {
   const stackRestartLifecycle = useRef<RestartLifecycleState>(createRestartLifecycleState());
   const stackRestartSuccessAnnounced = useRef(false);
   const stackStatusLoadRef = useRef<Promise<HomeLoadResult> | null>(null);
+  const sidebarRef = useRef<HTMLElement | null>(null);
   const [setupState, setSetupState] = useState<SetupState | null>(null);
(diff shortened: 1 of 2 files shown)


drkshrk:
This is better then what we had I think -> Not sure its the end all be all?  Open to suggestions....  I didn't personally like just makeing the Nav bar scrollable.

Red-Blink:
Thanks for working on this! After merging and testing it in the full Console, I reverted the change because it introduced a separate scrollbar inside the sidebar. The existing page scrolling already keeps the menu items reachable and provides a cleaner, more consistent experience without nested scrolling. No backend or other Console functionality was affected.

PR #156, opened 2026-08-14
docs/runtime: add collision-free multi-server SOP and host-port configurator

yacketrj:
## Summary

Adds a source-driven operator guide and configuration helper for running multiple independent Dune: Awakening battlegroups on isolated VMs sharing one public IPv4, plus one small runtime compatibility change.

## What's included

- **`runtime/scripts/multi-server-config.py`** — `plan`/`apply`/`verify` subcommands. Derives stock port defaults from repository source, applies a uniform `+1000`-per-instance offset to every managed host port/range, validates the complete port set for global collisions, and (on `apply`) backs up config, updates `.env`, and writes the authoritative UserEngine `Port`/`IGWPort` values.
- **`docs/runtime/MULTI-SERVER-SINGLE-PUBLIC-IP.md`** — the operator guide: port policy, the standard 3-instance profile table, NAT/firewall/hairpin setup, a phase-by-phase SOP (fresh config, verification, upgrade, rollback), a troubleshooting matrix, and the helper's own CLI reference.
- **`runtime/scripts/start-rabbitmq.sh`** — makes only the RabbitMQ game-management **host-side** port configurable (`RMQ_GAME_LOCAL_HTTP_PORT`, default unchanged at `15672`); no container-internal or protocol change.
- **`.env.example`** — documents the new advanced host-port override keys.
- **`docs/README.md`** — one new index entry for the guide above.

No other runtime behavior changes. Every default remains identical for an existing single-server install that never runs the new helper.

## Real multi-VM testing (per review feedback below)

Tested against two real, isolated VMs sharing one public IPv4 (not simulated): Player/Game + IGW traffic, hairpin NAT (verified via a live `tcpdump` capture), RabbitMQ/Director/Text Router/PostgreSQL/Console/Prometheus ports, a real upgrade-in-place (Instance 1 → 2 on an already-running server, zero data loss), and a real rollback to single-server (zero data loss). Full evidence: [testing comment](https://github.com/Red-Blink/dune-awakening-selfhost-docker/pull/156#issuecomment-5303375935).

That testing surfaced and fixed two real gaps in this branch:
- `apply()`'s running-container safety check blocked even after following the documented "stop the stack" procedure, because it didn't distinguish always-on management containers (console/orchestrator/probe, which hold no port this tool manages) from real game/database containers. Fixed with a narrow, tested exclusion.
- `plan`/`apply`'s collision validation only checks the ports generated in one invocation — it can't detect a collision against another, already-configured VM it wasn't asked about. Documented this limitation explicitly rather than leaving the "PASS" message ambiguous.

## Scope reduction (per review feedback below)

This PR's diff was **2,808 → 2,083 additions** (~26% smaller) after a dedicated pass: dropped a 530-line contributor-only staging runbook that added no value for a server owner, and cut ~200 lines of redundant per-VM repetition from the guide (duplicate tables, three near-identical command blocks where one representative example suffices, and generic hypervisor-sizing advice unrelated to this feature). Full breakdown: [scope-reduction comment](https://github.com/Red-Blink/dune-awakening-selfhost-docker/pull/156#issuecomment-5303745379). Nothing on this guide's own Maintainer Acceptance Checklist was cut.


Red-Blink:
Thanks for putting this together. Supporting multiple battlegroups behind one public IP is useful, and the collision-free port planning is heading in the right direction.

I’d like to keep this as a draft for now. The runtime change is relatively small, but the PR currently adds a large configurator and nearly 2,000 lines of documentation. Before merging, I’d like the scope simplified where possible and the complete workflow tested with real multi-VM deployments.

Please verify:

- Player/Game and IGW traffic for every instance
- RabbitMQ, Director, Text Router, PostgreSQL, Console, and optional Prometheus ports
- NAT, firewall, and hairpin behavior
- Fresh configuration and existing-server upgrades
- Re-running the helper without damaging manual configuration
- Rollback to the standard single-server layout
- Clear failures for collisions, invalid ranges, and exhausted ports

Once that is tested and the documentation is trimmed to what server owners actually need, we can review it again for merging.

yacketrj:
Real multi-VM testing completed against every item on the checklist above, run against an actual live deployment (two R740 VMs, one running stock Instance 1, one running Instance 2, sharing one public IPv4 through a real router with real NAT rules) -- not simulated.

## Player/Game and IGW traffic per instance

Confirmed both VMs' Player/Game UDP ports are correctly bound and reachable, including a rigorous hairpin-NAT test: captured real UDP packets sent to the public IP with `tcpdump` running on the target VM's own interface, confirming the router correctly reflects public->internal traffic for a LAN client:

```
$ tcpdump -ni ens18 udp port 8777 -c 5
10:02:16.243349 IP 192.168.68.127.49627 > 192.168.21.10.8777: UDP, length 23
[... 5 packets captured, matching 5 sent to the public IP ...]
```

IGW is intentionally *not* publicly forwarded on either VM (it's internal map-to-map traffic per the Web Console's own SetupWizard guidance) -- this is a deliberate, separately-tracked open question (`r740-dune-deployment-kit#84`), not an oversight in this testing.

## RabbitMQ, Director, Text Router, PostgreSQL, Console, Prometheus ports

All confirmed correctly assigned and listening on both VMs at their expected per-instance offsets (`dune status`'s Listeners section, both instances, all `OK`). RabbitMQ Game/Game-HTTP confirmed reachable from outside the LAN via a real router port-forward; the rest are correctly *not* publicly exposed (by design, per the guide's own NAT guidance) but confirmed bound and healthy locally.

## NAT, firewall, and hairpin behavior

NAT rules confirmed real and correct on the router for both VMs (Player/Game + RMQ ranges). Hairpin confirmed via the `tcpdump` capture above -- not just inferred from a lack of errors.

## Fresh configuration and existing-server upgrades

Tested a **real upgrade-in-place** on `dune-dev`: stopped the live stack, ran `apply --instance 2` on what was at that point a stock, already-running Instance 1 deployment, restarted, and confirmed full recovery -- `Overall: READY`, all 30 world partitions intact, Director/Population/FLS all `OK`. Zero data loss.

## Re-running the helper without damaging manual configuration

Deliberately injected a real mid-run failure (patched a bad value into the second `engine-set` call) and confirmed the resulting half-applied state (`.env` updated, `gameplay-profile.ini` only partially updated), then confirmed re-running the identical `apply` command afterward fully repairs the state (both `upsert_env()` and `engine-set` are idempotent). Full detail with real command output in `yacketrj/dune-awakening-selfhost-docker#280`'s PR body (a downstream feature built on top of this branch), under "Break/Fix -- What Happens if a Run Fails Mid-Execution."

## Rollback to the standard single-server layout

Tested a **real rollback** on `dune-dev`: stopped the live Instance 2 deployment, ran `apply --instance 1`, restarted, confirmed full recovery to stock ports (`7777`/`15432`/etc.), `Overall: READY`, all 30 world partitions intact. Then re-upgraded back to Instance 2 (see above) to leave the VM in its correct end state.

## Clear failures for collisions, invalid ranges, and exhausted ports

Confirmed the 33/34 instance-count boundary behaves as documented. Confirmed a real, local `.env` typo (a port set to the wrong instance's value) is caught cleanly by `verify` with a specific, actionable message.

**However, found and filed two real gaps during this testing**, both against the fork (not blocking this PR, but relevant to the "clear failures" ask):

- **`yacketrj/dune-awakening-selfhost-docker#282`**: `plan`/`apply`'s "Global collision validation: PASS" only ever validates the ports being generated *within one invocation* -- it has no way to detect a real collision against another, already-configured VM (confirmed by deliberately reproducing exactly that scenario). The docs' current phrasing ("compares every interval against every other interval across all generated VMs") could be read as checking a real network, which it does not. Suggest clarifying that wording.
- **`yacketrj/dune-awakening-selfhost-docker#283`**: following the documented "stop the stack, then apply" procedure literally still fails every time, because `dune stop` intentionally leaves the console/orchestrator/public-probe containers running (they're management tooling), but `apply`'s own running-container check treats them as blocking regardless. `--allow-running` is required in practice for the documented flow to ever succeed, not an edge case -- confirmed reproducibly on both directions of the real test above.

Neither finding blocks what's already correct in this PR -- both are suggestions for tightening the tool/docs before merge, filed as separate, trackable issues rather than left as review-comment-only observations.

yacketrj:
Addressed the scope/documentation-size half of the review comment above -- this PR's diff was actually growing (2,808 additions after the earlier fixes), not shrinking, so this is a real, separate pass focused specifically on that ask.

## What was cut

- **Removed `docs/runtime/MULTI-SERVER-SINGLE-PUBLIC-IP-UPSTREAM-PR.md` entirely** (530 lines). This file is a maintainer/contributor runbook for submitting *this exact PR* -- `git`/`gh` staging commands, not anything a server owner running the finished feature ever needs. The main guide's own "Intended Upstream File Scope" section already marked it optional.
- **Trimmed the main guide** (`MULTI-SERVER-SINGLE-PUBLIC-IP.md`, 1393 -> 1198 lines) by cutting genuine redundancy, not real content:
  - Removed Phase 7B "Manual configuration" (~100 lines) -- a full parallel VM1/VM2/VM3 walkthrough that just re-stated port values already shown earlier in the "Collision-Free Standard Profiles" table. Replaced with a short pointer.
  - Removed Phase 9's byte-for-byte duplicate of the earlier "Public NAT / Port-Forwarding Plan" tables.
  - Condensed Phase 10 (firewall) and Phase 21 (packet capture) from three near-identical per-VM command blocks to one or two representative examples plus a "same pattern for every other VM" note.
  - Removed Phase 26 "Capacity planning" (~20 lines of generic hypervisor sizing advice with no content specific to this feature).
  - Removed a 12-item flat file list that duplicated the already-present, more useful "Source-of-Truth Reference" table.

## What was deliberately kept

The core port-policy explanation, the authoritative port inventory, the Collision-Free Standard Profiles table, Configuration Sources, Addressing Model, the NAT/Port-Forwarding Plan (once, not duplicated), Phase 0's change-control checklist, the automated-helper walkthrough, hairpin validation, all verification phases, upgrade/rollback procedures, the Troubleshooting Matrix, the Automation Helper Reference, and the Maintainer Acceptance Checklist -- every one of these is real, non-redundant guidance a server owner or reviewer needs.

## Result

```
Before: 2,808 additions, 8 changed files
After:  2,083 additions, 7 changed files
```

A ~26% reduction. Verified nothing load-bearing was cut: re-ran `plan --instances 3/33`, the instance-34 rejection check, and the obsolete-value grep scan from this PR's own runbook after every edit -- all still pass identically, and every item on this PR's own Maintainer Acceptance Checklist still has real, present content backing it.

I recognize this is still a substantial PR -- the underlying feature (a globally collision-free port allocator across an arbitrary number of VMs, with `plan`/`apply`/`verify` subcommands) has real, irreducible complexity. Happy to make further scope cuts if there's a specific section you'd still like trimmed or moved elsewhere.

yacketrj:
## Follow-up on the two gaps found during live testing above

Both real gaps found during the multi-VM testing requested in your review comment are now resolved:

**#283 (real bug):** `apply()`'s running-container check blocked on the console/orchestrator/public-probe containers even after `dune stop` -- confirmed none of the three publish any host port this tool manages, so leaving them running is never actually unsafe. Fixed with a narrow, explicitly-tested exclusion (a new test confirms `dune-prometheus`, which DOES hold a managed port, still correctly blocks -- this isn't a blanket bypass).

**#282 (real limitation, docs-only fix):** cross-VM collision detection isn't possible with this tool's target topology (isolated VMs, no shared state -- tracked separately as future work in #278). Docs now state this limitation explicitly rather than leaving "Global collision validation: PASS" ambiguous.

5 new regression tests added (`runtime/scripts/test_multi_server_config.py`), all passing alongside the 3 pre-existing Python test suites (64 total). All 10 CI checks green. Both fixes and full reasoning are in commit `8b04d9a8`.

Red-Blink:
Thanks for the real multi-VM testing. I found several safety issues in the latest helper:

- `upsert_env()` explicitly writes `.env` as mode `0644`, which can expose the credentials stored there. It must preserve or enforce `0600`.

- `running_dune_containers()` fails open when Docker is unavailable or permission is denied, allowing configuration changes while the stack may still be running. This safety check must fail closed.

- `nat_lines()` still labels and prints the IGW range as part of the "Public NAT / port-forward plan," contradicting the stated decision not to forward IGW publicly.

- Apply is not transactional. If `.env` is changed and a later `usersettings.py` operation fails, the host is left with a mixed port configuration. Please restore the backup automatically on any failure or stage and validate all outputs before publishing them.

- The console uses host networking and does listen on `ADMIN_BIND_PORT`, so the explanation that it holds no rewritten host-facing port is inaccurate. Its running/staged behavior should be handled explicitly.

Please fix these paths and add failure-injection tests for Docker query failure, `usersettings` failure after `.env` changes, file permissions, and rollback.

yacketrj:
## Resolution -- addressed all five safety findings from the review above

Fixed in `89bc833f`:

1. **`upsert_env()` hardcoding mode 0o644**: now delegates to `atomic_write()`'s existing preserve-mode behavior (passing `None` instead of a hardcoded mode) instead of unconditionally loosening whatever permission `.env` already had -- matching `repair-host-runtime-permissions.sh`'s own `chmod --reference=.env` pattern for the same file. A brand-new `.env` still defaults to 644, matching this project's existing convention (`init.sh`/`manager.sh`/`memory.sh` all explicitly `chmod 644 .env`); this only stops an *existing*, tighter permission from being overwritten.

2. **`running_dune_containers()` failing open**: now raises `ConfigError` if Docker cannot be queried at all (daemon down, permission denied, docker missing), instead of returning `[]` and letting `apply()` proceed as if nothing were running. `command_apply()` only queries Docker when `--allow-running` is not already set, so the escape hatch for "I've already verified the stack is stopped" isn't blocked by this fail-closed behavior.

3. **`nat_lines()` printing the IGW range under the port-forward heading**: IGW is now listed separately from the real forward-target lines, clearly labeled `(informational, do NOT forward)` -- matching this project's own documented decision (and the live multi-VM testing evidence posted above) that IGW is internal map-to-map traffic and must never be publicly forwarded.

4. **Apply not being transactional**: added `restore_backup()` and wrapped the `.env` + `usersettings.py` write sequence in `command_apply()` in a try/except that rolls back to the exact pre-apply backup on any failure, raising a `ConfigError` that names the backup used. Either the command's net effect is the full new profile, or it's back to exactly what was there before -- never something in between.

5. **Console/`ADMIN_BIND_PORT` explanation being inaccurate**: corrected the doc -- the console does run with `network_mode: host` and does listen directly on `ADMIN_BIND_PORT`/`ADMIN_WEB_PORT`, a port `apply()` *does* rewrite. It's excluded from the running-container check because `apply()` only writes files and never touches already-running containers (so leaving it running during `apply` doesn't crash or corrupt anything) -- not because its port doesn't matter. The doc now says explicitly that the console must be restarted, like every other Dune service this tool reconfigures, before the new Admin Web port is actually live.

New regression tests for all five findings (`runtime/scripts/test_multi_server_config.py`) -- 13 tests total (was 5), including rewriting the one test that had explicitly asserted the old fail-open behavior into two tests asserting the new fail-closed contract.

Verified:
- `python3 runtime/scripts/test_multi_server_config.py`: 13 tests, all pass
- `python3 -c "ast.parse(...)"`: syntax valid
- Manual smoke test: `plan --instances 2` produces correct output, IGW correctly excluded from the forward-target lines
- `gitleaks detect`: 1 pre-existing finding in `docs/`, not in any file this commit touches
- `semgrep scan --config auto` on both changed Python files: 0 findings

All 10 CI checks passing on the updated head.

PR #157, opened 2026-08-14
fix: centralize port config, fix 6 places hardcoding stock ports instead of reading configured values

yacketrj:
## Summary

Fixes 6 places across the codebase that hardcode Instance-1 stock port
values instead of reading them from the already-configurable
`POSTGRES_PORT`/`RMQ_*_PORT`/`TEXT_ROUTER_PORT`/`DIRECTOR_PORT`/
`METRICS_PROMETHEUS_PORT`/`CLIENT_PORT_BASE`/`IGW_PORT_BASE` env vars
(confirmed already-merged upstream via `52d0abe6`, "Honor configurable
service ports" -- this PR does not introduce port configurability, it
finishes applying it consistently).

**Deliberately small and narrow in scope** -- this fixes exactly one
class of bug (hardcoded port values) with no dependency on any other
in-flight feature or draft PR. No secrets, no credentials, no
authentication paths, no schema changes, no upgrade/rollback story
beyond "this makes a display/instructional value correct instead of
wrong" -- every fallback default is byte-identical to today's hardcoded
value, so a deployment using stock ports sees zero behavior change.

## The bugs, in order of real-world impact

1. **`console/web/src/components/SetupWizard.tsx`** -- the setup
   wizard's own "Ports and Firewall" and "Review" steps show static
   port numbers regardless of what's actually configured. An operator
   running any non-default port configuration is told to forward the
   *wrong* ports during initial setup.
2. **`runtime/scripts/init.sh`** -- `dune init`'s "Public hosting
   reminder" has the identical bug (hardcoded `TCP 31982`/`31983`
   regardless of configured `RMQ_GAME_PORT`/`RMQ_GAME_HTTP_PORT`).
   Also adds the IGW UDP range, which this reminder omitted entirely
   regardless of port values.
3. **`console/web/src/components/PortChecklist.tsx`**,
   **`ReadinessTimeline.tsx`** -- port-number-to-friendly-label lookup
   tables hardcode all port keys, so a non-default deployment sees an
   unlabeled raw port number instead of "Postgres"/"RabbitMQ Admin"/etc
   in the console UI. Cosmetic, not functional.
4. **`console/api/src/server.js`**, **`console/web/src/api/client.ts`**
   -- `isPostgresUnavailableError()`/`friendlyApiError()`'s
   ECONNREFUSED-detection regex hardcodes the stock Postgres port.
   Low functional impact (the preceding generic `error.code ===
   "ECONNREFUSED"` / `connect ECONNREFUSED` checks already catch the
   real case) but wrong/misleading if ever relied upon or extended.

## The fix

Adds `resolvePorts()` to `console/api/src/config.js` as the single
source of truth for every host-facing port, following the exact same
env-var-with-stock-fallback pattern already correct in
`runtime/scripts/runtime-env.sh`'s `resolve_*_port()` functions and
`preflight.js`'s pre-existing (now refactored, not rewritten)
`configuredPorts()`. `db.js`, `server.js`, and `duneDb.js` now read
from this shared source instead of independently re-deriving stock
defaults inline.

Exposes the resolved values through the *existing*
`publicConfig()` -> `GET /api/auth/state` -> frontend flow -- `App.tsx`
already fetches this response on every page load; this just adds a
field it previously received and discarded. New
`console/web/src/api/serverPorts.ts` is the frontend-side cache other
components read from, mirroring `config.js`'s role on the backend.

## Root cause

The port-configurability feature (`52d0abe6`) correctly made these
values configurable at the resolution layer, but no process step at
that time audited every *other* place a port number appears in the
codebase. This PR is that audit, plus the fix -- found via a live
multi-instance deployment surfacing `server.js`'s hardcoded value in
practice, then systematically grepped for every stock port value
across `console/api/src`, `console/web/src`, and `runtime/scripts`.

## Testing

```
console/api: node --test test/config.test.js -- 4/4 pass (3 new
  regression tests added: resolvePorts() with configured ports,
  resolvePorts() stock-value fallback, publicConfig() exposing ports)
console/api: npm test (full suite) -- 995/1017 pass. Confirmed via a
  clean upstream/main worktree that the 7 failing tests are
  pre-existing (require live DB/metrics-stack infrastructure not
  available locally) and identical before/after this change --
  992/1014 on clean main, same 7 failures by name.

console/web: npx tsc --noEmit -- clean
console/web: npx vitest run -- 378/378 pass (0 pre-existing failures)
console/web: npm run build -- succeeds

runtime/scripts/init.sh: bash -n -- clean
runtime/scripts/init.sh: shellcheck -x -S warning -- clean on every
  line touched (3 pre-existing SC2155 warnings elsewhere in the file,
  confirmed outside this diff's line ranges)

gitleaks detect + semgrep --config p/default on all 13 changed files:
  0 findings in either
```

## Explicitly out of scope

- No change to the underlying port-configurability mechanism itself
  (`runtime-env.sh`'s resolvers, `.env` variable names) -- this PR only
  fixes secondary consumers that failed to read from it.
- Does not touch anything related to secrets management, credentials,
  or authentication.

### DIFF
--- a/console/api/src/config.js
+++ b/console/api/src/config.js
@@ -48,12 +48,25 @@ export function resolvePorts(env = process.env, repoRoot = process.cwd()) {
   // (e.g. Port=0 or a 10-digit garbage number) that bypasses range
   // validation entirely and flows through to the frontend and the
   // public-hosting reminder text unvalidated.
-  const clientBase = enginePorts.port !== null
+  const rawClientBase = enginePorts.port !== null
     ? portValue(enginePorts.port, 7777)
     : (legacyEnginePorts.port !== null ? portValue(legacyEnginePorts.port, 7777) : 7777);
-  const igwBase = enginePorts.igwPort !== null
+  const rawIgwBase = enginePorts.igwPort !== null
     ? portValue(enginePorts.igwPort, 7888)
     : (legacyEnginePorts.igwPort !== null ? portValue(legacyEnginePorts.igwPort, 7888) : 7888);
+  // Upstream review finding: portValue() only range-checks the base
+  // port itself (1-65535), but every Player/Game and IGW base is
+  // actually the start of a 34-slot range (see spawn-server.sh's
+  // game_end=$((CLIENT_PORT_BASE + 33))/igw_end=$((IGW_PORT_BASE + 33)),
+  // covering the maximum 34 world partitions this project supports) --
+  // a base in the valid 1-65535 range can still overflow past 65535
+  // once the +33 offset for the highest partition (or the console's
+  // own +1 "secondary" port, used for the stock 2-partition default) is
+  // applied. A base that would overflow is exactly as unusable as one
+  // that's already out of range, so it falls back to stock the same
+  // way portValue() already does for a directly out-of-range value.
+  const clientBase = rawClientBase + 33 <= 65535 ? rawClientBase : 7777;
+  const igwBase = rawIgwBase + 33 <= 65535 ? rawIgwBase : 7888;
   return {
     postgres: portValue(env.POSTGRES_PORT || env.DUNE_DB_PORT || env.PGPORT, 15432),
     rmqAdmin: portValue(env.RMQ_ADMIN_PORT, 32573),
@@ -104,7 +117,19 @@ function readEnginePortsFromProfile(repoRoot) {
   // for hand-edited/out-of-band files, not the common path.
   const normalized = text.replace(/\r\n/g, "\n");
   const sectionMatch = normalized.match(/^\[Engine:URL\]\s*$([\s\S]*?)(?=^\[|\s*$(?!\n))/m);
-  const sectionText = sectionMatch ? sectionMatch[1] : normalized;
+  // Upstream review finding: if [Engine:URL] is absent entirely (a
+  // stripped-down or hand-edited profile file), this previously fell
+  // back to scanning the WHOLE file for a Port=/IGWPort= match -- which
+  // could silently pick up an unrelated key from a different section
+  // (e.g. [Engine:ConsoleVariables]'s own Port= override, which means
+  // something entirely different) and report it as the real game port.
+  // The real Python tool (usersettings.py's profile_get_key()) only
+  // ever looks inside the named section; if the section doesn't exist,
+  // it returns nothing for that key, not a value from elsewhere in the
+  // file. Match that exactly: no section match -> no port match, fall
+  // through to the legacy usersettings.json / stock fallback chain.
+  if (!sectionMatch) return { port: null, igwPort: null };
+  const sectionText = sectionMatch[1];
   // Layer 3 audit regression: if a section has a duplicate key (Port=
   // appearing twice, which usersettings.py's own
   // _advanced_editor_duplicate_key_warnings() explicitly anticipates and
--- a/console/api/src/config.js
+++ b/console/api/src/config.js
@@ -14,9 +14,25 @@ export const APP_NAME = "Dune Docker Console";
 // places across the codebase that hardcoded these stock values directly
 // instead of reading them from a shared source, breaking any deployment
 // running non-default configured ports).
-export function resolvePorts(env = process.env) {
-  const clientBase = portValue(env.CLIENT_PORT_BASE, 7777);
-  const igwBase = portValue(env.IGW_PORT_BASE, 7888);
+export function resolvePorts(env = process.env, repoRoot = process.cwd()) {
+  const enginePorts = readEnginePortsFromProfile(repoRoot);
+  // Player/Game and IGW base ports are authoritative in
+  // runtime/generated/gameplay-profile.ini's [Engine:URL] section
+  // (written via runtime/scripts/usersettings.py engine-set, e.g. by
+  // the Maps UI or multi-server-config.py) -- NOT env vars. .env's
+  // CLIENT_PORT_BASE/IGW_PORT_BASE are documented as
+  // "compatibility/console metadata" only (see
+  // docs/runtime/MULTI-SERVER-SINGLE-PUBLIC-IP.md): a deployment that
+  // changed these via the Maps UI directly, without also re-running
+  // multi-server-config.py, would have a stale .env value while the
+  // real game server already uses the new one -- reading the profile
+  // file first avoids exactly that staleness. .env is kept as a
+  // secondary fallback (useful before the profile file exists at all,
+  // e.g. immediately after multi-server-config.py writes .env but
+  // before its own engine-set call has run), and the stock literal is
+  // the final fallback.
+  const clientBase = enginePorts.port ?? portValue(env.CLIENT_PORT_BASE, 7777);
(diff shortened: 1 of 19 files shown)


yacketrj:
## Reverting to draft — process correction

I marked this ready for review without completing the layered eight-hats
audit (design/implementation/integration) this account's own operating
rules require before an upstream PR can leave draft. I substituted
passing tests, linting, and security scans for that audit, which is not
the same thing — real CI passing is verification that the code works,
not that it has been independently reviewed from every angle (security,
architecture, GRC, network, cloud, UI, DBA, QA) the way a change heading
to a third-party maintainer is supposed to be.

Putting this back in draft. I'll run the actual layered audit against
this diff and only bring it back to ready based on what that review
finds — not on an assumption that it'll come back clean. Apologies for
the noise; wanted to be upfront about this rather than quietly leave it
as-is.

yacketrj:
## Resolution — layered eight-hats audit completed

Following up on the process-correction comment above: the full layered audit (design/implementation/integration) was run against this PR, not assumed clean. Findings and fixes, in order:

**Layer 1 (design)** — 1 CRITICAL: \`resolvePorts()\`'s \`clientBase\`/\`igwBase\` read from \`.env\`'s \`CLIENT_PORT_BASE\`/\`IGW_PORT_BASE\`, which are documented as "compatibility/console metadata" only. The real, authoritative source is \`runtime/generated/gameplay-profile.ini\`'s \`[Engine:URL]\` section, written via \`usersettings.py engine-set\` (e.g. by the Maps UI). An operator changing the port via the Maps UI without also re-running the multi-server config tool would see a stale value in the console UI while the real game server already used the new one -- silently, with no error. Verified live against two real deployments. Fixed in \`da8036aa\`.

**Layer 2 (implementation)** — 1 HIGH, 2 MEDIUM: \`config.ports\` was captured once at process startup and never re-evaluated, silently undercutting the Layer 1 fix (correct once at boot, stale forever after the first live port change) -- now a live getter. A CRLF-encoding edge case in the profile parser could silently drop a key. 3 of 6 \`resolvePorts()\` call sites relied on \`process.cwd()\` coincidentally matching the real repo root. All fixed in \`a2d491d7\`.

**Layer 3 (integration)** — 2 CRITICAL: found and directly reproduced two real divergences between the Node-side port resolver and the shell-side resolver (\`runtime-env.sh\`, backed by \`usersettings.py\`) that this file's own comment claims "must never drift." Neither Layer 1 nor Layer 2 caught these because each audited one side in isolation; this layer diffed both real implementations against identical inputs and found they disagreed in two concrete, reachable scenarios. Fixed in \`9b3534fb\`, with new tests that run both real implementations side-by-side so this can't silently regress.

Current state: all 10 CI checks passing, \`mergeStateStatus: CLEAN\`, still correctly in draft pending final review. Full detail for each layer is in the respective commit messages linked above.

Red-Blink:
The centralized port resolution is a good improvement, but the latest head still has consistency problems:

- The frontend port cache is populated only once by the initial `/api/auth/state` request. If `Port` or `IGWPort` is changed from the Maps UI, the backend getter updates but the browser cache remains stale until the page is reloaded. Please refresh the port state after relevant saves or expose it through a reactive context/hook.

- If `[Engine:URL]` is absent, `readEnginePortsFromProfile()` scans the entire INI and can accidentally accept a `Port=` key from another section. It should return null values when that section is missing.

- `resolvePorts()` validates the base port but not derived values such as `clientBase + 33` or secondary ports, which can exceed `65535`.

- PostgreSQL precedence still differs: `resolvePorts()` prefers `POSTGRES_PORT`, while `discoverDbConfig()` prefers `DUNE_DB_PORT/PGPORT`. If more than one is set, status/preflight can disagree with the actual database connection.

Please align these paths and add regression tests for live updates, a missing `[Engine:URL]` section, conflicting PostgreSQL variables, and upper-bound port ranges.

yacketrj:
Thanks for the detailed follow-up -- all four addressed:

**Stale frontend cache:** added `refreshServerPorts()`, called after every raw UserEngine.ini save from the Maps UI (the only console-driven write path for `Port`/`IGWPort`), so the frontend cache no longer requires a full page reload to pick up a change.

**Missing `[Engine:URL]` section:** `readEnginePortsFromProfile()` now returns null for both fields when the section itself is absent, instead of falling back to scanning the whole file -- matches `usersettings.py`'s own `profile_get_key()`, which only ever looks inside the named section.

**Upper-bound validation:** `resolvePorts()` now also validates the derived 34-slot range each base anchors (`spawn-server.sh`'s `game_end = base + 33`, covering the max 34 world partitions this project supports), not just the base port itself -- falls back to stock the same way an already-out-of-range base does.

**PostgreSQL precedence:** `discoverDbConfig()` no longer implements its own precedence (`DUNE_DB_PORT || PGPORT || ...`) -- it now always delegates to `resolvePorts()` for this field, so the two paths can no longer disagree.

New regression tests for all four. Re-verified directly (not just trusting the commit message): `node --test` 295/295 pass, `npx vitest run` 6/6 pass on the new test file, `tsc --noEmit` clean, `npm run build` succeeds. CI is green on this branch.

yacketrj:
## Resolution -- addressed all four findings from the review above

Fixed in `a6fee755`:

1. **Frontend port cache staleness**: added `refreshServerPorts()` (`api/serverPorts.ts`) and call it from `MapsPanel.tsx`'s raw UserEngine.ini save path -- the only console-driven write path for Port/IGWPort (the structured per-field editor deliberately excludes them). PortChecklist/ReadinessTimeline/SetupWizard now pick up a live port change without a page reload.

2. **`readEnginePortsFromProfile()` reading from the wrong section**: if `[Engine:URL]` is absent, this now returns `{ port: null, igwPort: null }` instead of falling back to scanning the entire file -- matching `usersettings.py`'s `profile_get_key()`, which only ever looks inside the named section.

3. **Missing upper-bound validation on derived ports**: `resolvePorts()` now rejects a Port/IGWPort base whose `+33`-slot partition range (see `spawn-server.sh`'s `game_end=$((CLIENT_PORT_BASE + 33))`, covering the max 34 world partitions this project supports) would overflow past 65535, falling back to stock the same way an already-out-of-range base does.

4. **PostgreSQL port precedence divergence**: `discoverDbConfig()` now always delegates to `resolvePorts()` for the Postgres port instead of re-implementing a separate precedence chain, so there's exactly one place this logic can ever drift from itself.

New regression tests for all four findings (`console/api/test/config.test.js`, `db.test.js`, `console/web/src/api/serverPorts.test.ts`).

Verified:
- `node --test console/api/test/*.test.js`: 1035 tests, 1015 pass, 5 fail (pre-existing live-DB-dependent tests -- identical failure set with and without this change, confirmed via `git stash` comparison)
- `npx vitest run` (console/web): 49 files, 386 tests, all pass
- `npx tsc --noEmit` (console/web): clean
- `npm run build` (console/web): succeeds
- `gitleaks detect`: 1 pre-existing finding in `docs/`, not in any file this commit touches
- `semgrep scan --config auto` on all 4 changed source files: 0 findings

All 10 CI checks passing on the updated head.

PR #171, opened 2026-08-18
feat(discord): add command catalog endpoint (GET /api/integrations/discord/catalog)

yacketrj:
# Discord command catalog endpoint (`GET /api/integrations/discord/catalog`)

## Executive Summary

The companion Discord bot (a separate, community-maintained project) has to independently track which adapter routes are live, what capability each one needs, and what Discord-facing metadata to register for each one. Today that happens by hand — someone reads this repo's `adapter.js`/`policy.js`/`routes.js`, then manually updates the bot's own copy. This PR adds a single, read-only, machine-readable endpoint (`GET /api/integrations/discord/catalog`) that gives any bot integration that same information directly, mechanically kept in sync with this repo's real route table.

This is Phase 1 of a 4-phase design already merged upstream, docs-only, as `docs/rfc-command-discovery.md`. Phases 2–4 (bot-side generator script, bot runtime consumption, dynamic refresh/autocomplete) are separate, future work and explicitly out of scope here. This PR is purely additive — no existing route, field, or response shape is renamed, removed, or changed in a breaking way.

## What Changed and Why

- **New** `console/api/src/integrations/discord/commandCatalog.js` — `buildCommandCatalog()` composes a catalog from `DISCORD_LIVE_ADAPTER_ROUTES`, `DISCORD_CAPABILITIES`, and a new `policy.js` export, `minTierForCapability()`, plus a per-route metadata table (name, description, param shape) that is genuinely new information — it does not exist as code anywhere in this repo today, since command metadata currently only exists inside the separate bot repo's own source.
- **New** `GET /api/integrations/discord/catalog` (`routes.js`) — bearer-token auth only, matching the existing `/health` endpoint's auth model. No per-actor capability check: this returns metadata about route shape, not game or player data.
- `/health` gains a `protocolVersion` field, for future version-negotiation if this catalog's contract ever needs a breaking change.
- **New** `policy.js` export, `minTierForCapability(capability)` — derives "the minimum role tier that grants this capability" directly from the existing `CAPABILITY_BY_TIER` table, so `commandCatalog.js` never has to hand-maintain a second, parallel copy of that mapping that could silently drift from the real one.
- `docs/console/API-REFERENCE.md` — added the new route to the existing Discord Adapter table.

**Why this design, not a hand-maintained JSON file:** an earlier, simpler approach (a static JSON file listing routes/capabilities, authored once and manually kept up to date) was considered and rejected before this PR was written, for the same reason the bot's own current `LIVE_ROUTES`/`PLANNED_ROUTES`/`UNMERGED_ROUTES`/`MISSING_ROUTES` classification is the problem this PR exists to fix: a hand-maintained artifact drifts. `buildCommandCatalog()` instead asserts, on every call, that its metadata table covers `DISCORD_LIVE_ADAPTER_ROUTES` exactly — no live route missing an entry, no entry left behind for a retired route — and throws if either is violated, so route-coverage drift is structurally impossible rather than merely discouraged.

## Testing

Real, actual command output, run against this exact branch:

```
$ cd console/api && npm test
...
1..1156
# tests 1162
# suites 0
# pass 1162
# fail 0
# cancelled 0
# skipped 0
# todo 0
```

```
$ node --test test/discordCommandCatalog.test.js test/discordAdapter.test.js test/discordPolicy.test.js
1..N
# pass (all green, 0 failures)
```

New tests added by this PR (`discordCommandCatalog.test.js`, 13 cases) specifically verify: every live route has exactly one catalog entry and vice versa (with two dedicated tests that construct a genuinely mismatched route/metadata pair and assert the coverage-check throw actually fires, not just re-asserting the non-drifted case); every `minTier` is drawn from a real tier in `DISCORD_ROLE_TIERS`; self-scoped capabilities are all marked `selfScoped`; `broadcast` is the only write-gated subcommand; the catalog's own version matches the exported constant; the new `CATALOG` route is deliberately excluded from `DISCORD_LIVE_ADAPTER_ROUTES` (metadata about routes is not itself a data route); the known `routeEnforcesCapability: false` exceptions (`HEALTH`, `BACKUPS_LIST`, `VERSION`) are exactly and only those three; and no two catalog entries share a verbatim-identical description (guards against an accidental copy-paste of one entry's description into another).

`minTierForCapability()`'s correctness is independently cross-checked in `discordPolicy.test.js` against `discordActorCan()` for every real capability — not just re-reading `CAPABILITY_BY_TIER`'s shape a second time — so a future capability reassignment between tiers would fail a real test if this function's answer ever stopped matching real enforcement.

## Fresh Install

No change to fresh-install behavior. This is a new, always-on but purely read-only, additive endpoint; an operator who never has a bot query `/catalog` sees no behavior change at all. Nothing is written to disk by this feature — `buildCommandCatalog()` operates entirely on already-frozen, in-memory module constants (no database access, no filesystem access, no environment-variable input at all).

## Migration / Upgrade Path

No migration required. This PR adds no new required environment variable, config key, or schema, and changes no existing response field's meaning. `/health`'s new `protocolVersion` field is additive — an existing bot that doesn't know about it simply ignores it. An operator who pulls this update and whose bot never calls the new endpoint will not notice this PR exists.

## Break/Fix — What Happens if a Run Fails Mid-Execution

This endpoint has no multi-step execution to fail "mid-run" in the way a stateful operation (a backup, a migration) would — `buildCommandCatalog()` is a single, synchronous, side-effect-free computation over frozen constants, with no I/O, no partial state, and no cleanup path required. The two failure modes that do exist are both deliberately tested:

- **A live route is added to `DISCORD_LIVE_ADAPTER_ROUTES` without a matching `COMMAND_METADATA` entry** (or vice versa, a stale entry left behind after a route is retired): `buildCommandCatalog()` throws synchronously, on every call, rather than silently returning an incomplete or stale catalog. Covered by two dedicated tests that construct exactly this scenario and assert the throw. A caller (the route handler in `routes.js`) that hits this throw will surface a real 500 to the requesting bot rather than a silently wrong 200 — a bot developer will see the failure immediately rather than trust bad data.
- **The bearer token is missing or wrong:** the existing `requireDiscordBotToken()` gate (unchanged by this PR, shared with `/health`) rejects the request with 401/503 before any route-specific code runs at all — confirmed by direct read of the dispatch order in `routes.js` (the token check is unconditional and executes before every path-specific branch, including this new one) and by the existing end-to-end HTTP test in `discordAdapter.test.js`.

## What if the Bearer Token Is Lost

Not a new concern introduced by this PR — this endpoint reuses the exact same, already-existing bearer-token credential every other Discord adapter route (`/health`, `/status`, etc.) already requires. This PR introduces no new credential, no new secret, and no new recovery/rotation surface. The existing token's own generation/rotation procedure (unchanged, documented elsewhere in this repo) applies identically here.

## Eight Hats Audit Summary

This feature went through all three audit layers this contributor's own review discipline requires before an upstream PR is opened:

**Layer 1 (design)** — completed and recorded in the already-merged `docs/rfc-command-discovery.md` ("Eight-Hat Layer 1 completed. Findings incorporated.").

**Layer 2 (implementation)**, against the actual code (dispatched as independent reviewers per hat, verifying claims against real code/command output rather than reasoning abstractly, not a single combined pass) — found and fixed, before this PR was opened:
- **CRITICAL (QA):** a test named "catalog throws if a live route is missing metadata (simulated drift)" never actually simulated drift — it re-imported the same, consistent module and asserted non-throw, so the actual throw path had zero test coverage. Fixed: `buildCommandCatalog()` now accepts optional `liveRoutes`/`metadata` parameters (defaulting to real production data on every real call site) specifically so tests can inject a genuinely mismatched pair; two new tests construct missing-metadata and stale-metadata scenarios and assert the exact throw fires.
- **HIGH (QA):** no negative-path test existed for route retirement (stale metadata left behind) — closed by the same two tests above.
- **HIGH (GRC):** the new route's unusual (bearer-token-only, no per-capability check) auth model was not documented anywhere a compliance reviewer would look first — added an explicit row to `docs/console/API-REFERENCE.md` calling out the exception and why.
- **MEDIUM (Architect/Security):** the minimum-tier-per-capability mapping was originally a hand-maintained shadow of `policy.js`'s real `CAPABILITY_BY_TIER`, correct at the time but with zero automated protection against future drift. Closed by exporting `minTierForCapability()` from `policy.js` itself (deriving the answer from the real table directly) and adding 3 tests in `discordPolicy.test.js` that independently cross-check every capability's claimed minimum tier against `discordActorCan()`.
- **MEDIUM (Network):** `buildCommandCatalog()` recomputed its coverage assertion and group composition on every single request with no caching. Since the underlying data is frozen and load-time-constant, the zero-argument (production) call path is now memoized after its first call; memoization is explicitly skipped whenever a test-injection argument is passed, verified by dedicated tests confirming injected calls never share the production cache.
- Findings not applicable/deferred (adapter-wide, pre-existing, out of this PR's specific blast radius, not specific to this new endpoint): absence of rate limiting on every bearer-token-only Discord adapter GET route (`/health`, `/version`, `/backups/list`, and now `/catalog`) — this PR's endpoint inherits exactly the same posture `/health` already has, rather than diverging from it; a documentation gap in an unrelated admin guide's TLS/tunnel guidance for cross-host bot↔Core traffic.

**Layer 3 (integration)**, against this PR's complete, final 8-file diff as a whole (not each file reviewed in isolation) — found and fixed:
- **MEDIUM (Architect/Security/QA):** the `HEALTH` catalog entry declared a capability (`STATUS_READ`) with no `routeEnforcesCapability: false` override, defaulting to "enforced" — but the real route handler, `discordAdapterHealth()`, never calls a capability check at all, identical in enforcement posture to `BACKUPS_LIST`/`VERSION`, both of which already correctly carry that flag. Independently re-derived the real capability+enforcement for all 32 live routes directly from `routes.js`/`adapter.js` source (not from the catalog's own claims) and cross-checked against the table — this was the only mismatch. Impact was low (an over-restrictive false positive for a future bot-side permissions UI, not an under-protection), but exactly the kind of "hand-authored field within an entry is not drift-proof" gap this file's own docblock already warns about. **Fixed**, with the existing coverage test updated to include `HEALTH` in the known-exceptions set.
- Cross-file consistency (the new `minTierForCapability()` export, the `/catalog` route's placement relative to the existing bearer-token auth gate, the new `docs/console/API-REFERENCE.md` row against the real implementation) — all independently re-verified against this PR's exact diff and confirmed correct.
- No secret/credential exposure risk: `buildCommandCatalog()` takes no `config`, `req`, or environment-derived input at all, so there is no code path through which a token or config value could reach the response.

Every CRITICAL/HIGH finding from Layers 1–3 is resolved in the code as submitted in this PR; the one Layer 3 MEDIUM finding above is also resolved, not deferred.

## Security Checks

Run directly against this branch:

```
$ cd console/api && npm test
1162/1162 pass, 0 fail

$ gitleaks protect --staged
no leaks found

$ semgrep --config auto <7 changed JS files>
Ran 200 rules on 7 files: 0 findings
```

No new dependency, no new network exposure (this endpoint is a pure in-process addition to an existing HTTP dispatcher — no new port, no new outbound call, no new DNS dependency), and no IAM/cloud-provider footprint of any kind.

## Live-Deployment Validation

Run against a real, running deployment (a dev/staging instance, not just
the automated test suite above), immediately before this PR was marked
ready for review:

1. Backed up the three modified files and the running container's
   pre-existing `/app` copies of them, plus the current `.env`.
2. Overlaid this branch's 4 changed/new files into the running console
   container (`docker cp`), enabled the adapter with a fresh, throwaway
   bearer token (`DUNE_DISCORD_ADAPTER_ENABLED=true` +
   `DUNE_DISCORD_ADAPTER_TOKEN=<random 64-char hex, generated for this
   test only>`), and restarted the container process.
3. Confirmed all 11 other containers on the host (game servers, database,
   RabbitMQ, orchestrator, autoscaler) were unaffected throughout —
   `dune status` reported `Overall: READY` before, during, and after this
   test, population unchanged (0 online at the time).

Real, actual responses from the live instance:

```
$ curl -s http://localhost:9088/api/integrations/discord/health -H "Authorization: Bearer <token>"
{"ok":true, ..., "protocolVersion":1, ...}

$ curl -s -o /dev/null -w '%{http_code}' http://localhost:9088/api/integrations/discord/catalog -H "Authorization: Bearer <token>"
200
$ curl -s http://localhost:9088/api/integrations/discord/catalog -H "Authorization: Bearer <token>"
{"ok":true,"protocolVersion":1,"catalog":{"version":1,"groups":[... all 8 groups, 32 routes, correct minTier/capability/routeEnforcesCapability per route ...]}}

$ curl -s -o /dev/null -w '%{http_code}' http://localhost:9088/api/integrations/discord/catalog
401
$ curl -s http://localhost:9088/api/integrations/discord/catalog
{"ok":false,"code":"missing_bot_token","error":"Missing adapter credential."}

$ curl -s http://localhost:9088/api/integrations/discord/catalog -H "Authorization: Bearer wrong-token"
{"ok":false,"code":"invalid_bot_token","error":"Invalid adapter credential."}
(401)

$ curl -s -X POST http://localhost:9088/api/integrations/discord/catalog -H "Authorization: Bearer <token>"
{"ok":false,"code":"not_found","error":"Discord adapter route not found."}
(404 -- no POST handler for this route, as expected)

$ curl -s http://localhost:9088/api/integrations/discord/version -H "Authorization: Bearer <token>"
{"ok":true,"version":"v1.3.88"}
```

The live `HEALTH` catalog entry correctly shows `"routeEnforcesCapability":false`
— confirming the Layer 3 fix (see above) is genuinely correct against a
real running instance, not just the test suite.

After validation, fully reverted: restored the container's original
`/app` files (verified via matching checksums against pre-test backups),
removed the newly-added file, restored the original `.env` exactly
(verified via `diff`), and recreated the container so it picked up a
clean copy from the unmodified image. Confirmed post-revert: adapter
reports disabled again, throwaway token no longer present anywhere, all
containers' status unchanged from pre-test, `dune status` still `READY`.

## Rollback Safety

Purely additive with no schema/state to roll back. Reverting this PR's commit removes the new endpoint and the `protocolVersion` field; nothing else changes behavior. No data migration, no cleanup step, no operator action required either direction.

## Files Changed

- `console/api/src/integrations/discord/adapter.js` — new `DISCORD_ADAPTER_ROUTES.CATALOG` constant, new `DISCORD_CATALOG_PROTOCOL_VERSION` export, `/health` gains `protocolVersion`.
- `console/api/src/integrations/discord/commandCatalog.js` — new file, `buildCommandCatalog()`.
- `console/api/src/integrations/discord/policy.js` — new `minTierForCapability()` export.
- `console/api/src/integrations/discord/routes.js` — new `GET /api/integrations/discord/catalog` dispatch entry.
- `console/api/test/discordAdapter.test.js`, `console/api/test/discordCommandCatalog.test.js` (new), `console/api/test/discordPolicy.test.js` — test coverage described above.
- `docs/console/API-REFERENCE.md` — new route documented.

## Not in Scope for This PR

- Any change to the companion Discord bot repo (a separate, future PR will consume this endpoint there).
- Per-guild catalogs, dynamic/background refresh, or autocomplete infrastructure (Phase 4 of the design).
- Any change to write-command routes or authorization (this is a read-only metadata endpoint about the *existing*, already-shipped read-only route surface).
- Rate limiting for bearer-token-only adapter GET routes generally (pre-existing, adapter-wide gap noted above, not specific to this endpoint).

### DIFF
--- a/console/api/src/integrations/discord/adapter.js
+++ b/console/api/src/integrations/discord/adapter.js
@@ -124,7 +124,15 @@ export function validateDiscordActor(actorPayload) {
 // its meaning changes) -- NOT bumped for ordinary route additions, which
 // are additive and always safe for a bot that doesn't yet know about them.
 // See commandCatalog.js and docs/rfc-command-discovery.md §3.4.
-export const DISCORD_CATALOG_PROTOCOL_VERSION = 1;
+//
+// Bumped 1 -> 2 alongside commandCatalog.js's own CATALOG_VERSION: a
+// subcommand's `route` (a single string) became `routes` (an array),
+// needed to correctly represent 3 (group, subcommand) pairs that
+// genuinely fan out to two backing adapter routes each (upstream PR #171
+// review). No live consumer exists yet to break (Phase 2/3 bot-side
+// generator is still unimplemented), so this is forward hygiene per this
+// comment's own stated policy, not a fix for an active breakage.
+export const DISCORD_CATALOG_PROTOCOL_VERSION = 2;
 
 export async function discordAdapterHealth(config) {
   return {
--- a/console/api/src/integrations/discord/adapter.js
+++ b/console/api/src/integrations/discord/adapter.js
@@ -37,7 +37,16 @@ export const DISCORD_ADAPTER_ROUTES = Object.freeze({
   VERSION: "/api/integrations/discord/version",
   SERVERS: "/api/integrations/discord/servers",
   PORTS: "/api/integrations/discord/ports",
-  DB: "/api/integrations/discord/db"
+  DB: "/api/integrations/discord/db",
+  // CATALOG is deliberately NOT added to DISCORD_LIVE_ADAPTER_ROUTES below.
+  // It is metadata ABOUT the live routes, not itself one of them -- adding
+  // it there would require commandCatalog.js's COMMAND_METADATA to have an
+  // entry describing itself, which is circular and not meaningful (a
+  // catalog entry for "the catalog"). Its own liveness/auth is handled
+  // directly by its route.js dispatch entry (bearer-token auth only, same
+  // as HEALTH) rather than the capability/tier system, since it is
+  // read-only metadata about route shape, not game or player data.
+  CATALOG: "/api/integrations/discord/catalog"
 });
 
 export const DISCORD_LIVE_ADAPTER_ROUTES = Object.freeze([
@@ -110,6 +119,13 @@ export function validateDiscordActor(actorPayload) {
   return normalizeDiscordActor(actorPayload);
 }
 
+// Bumped only when the command-catalog contract itself changes in a way
+// that could break an older bot's assumptions (e.g. a field is removed or
+// its meaning changes) -- NOT bumped for ordinary route additions, which
+// are additive and always safe for a bot that doesn't yet know about them.
+// See commandCatalog.js and docs/rfc-command-discovery.md §3.4.
+export const DISCORD_CATALOG_PROTOCOL_VERSION = 1;
+
 export async function discordAdapterHealth(config) {
   return {
     ok: true,
@@ -123,7 +139,8 @@ export async function discordAdapterHealth(config) {
     routes: DISCORD_LIVE_ADAPTER_ROUTES,
     liveRoutes: DISCORD_LIVE_ADAPTER_ROUTES,
     plannedRoutes: DISCORD_PLANNED_ADAPTER_ROUTES,
-    rolePolicy: discordRolePolicyHealth()
+    rolePolicy: discordRolePolicyHealth(),
+    protocolVersion: DISCORD_CATALOG_PROTOCOL_VERSION
   };
 }
 
--- a/console/api/src/integrations/discord/commandCatalog.js
+++ b/console/api/src/integrations/discord/commandCatalog.js
@@ -40,7 +40,18 @@
 import { DISCORD_ADAPTER_ROUTES, DISCORD_LIVE_ADAPTER_ROUTES } from "./adapter.js";
 import { DISCORD_CAPABILITIES, minTierForCapability } from "./policy.js";
 
-export const CATALOG_VERSION = 1;
+// Bumped from 1 to 2 for this revision: subcommand.route (a single string)
+// became subcommand.routes (an array) for every (group, subcommand) pair
+// that fans out to more than one backing adapter route -- a breaking shape
+// change for any future consumer that assumed one route per subcommand, per
+// this constant's own original design intent (see adapter.js's
+// DISCORD_CATALOG_PROTOCOL_VERSION comment: bump on "a field is removed or
+// its meaning changes", not on ordinary additions). No live consumer exists
+// yet (Phase 2/3 bot-side generator is still unimplemented -- confirmed via
+// grep across arrakis-control-panel's src/ for buildCommandCatalog/
+// DISCORD_CATALOG_PROTOCOL_VERSION, zero hits), so this bump is forward
+// hygiene, not a fix for an active breakage.
+export const CATALOG_VERSION = 2;
 
 // One entry per DISCORD_LIVE_ADAPTER_ROUTES member. Keys are route path
 // strings (DISCORD_ADAPTER_ROUTES.* values), not route constant names, so
@@ -51,7 +62,43 @@ export const CATALOG_VERSION = 1;
 // checks for this route (verified by direct reading of both files, not
 // inferred) -- diagnostic-mode STATUS uses LOGS_READ instead of
 // STATUS_READ; that distinction is preserved here via diagnosticCapability.
-const COMMAND_METADATA = Object.freeze({
+//
+// method: the exact HTTP method routes.js's real dispatch table checks for
+// this route (`req.method === "GET"|"POST"`) -- defaults to "POST" when
+// omitted (buildCommandCatalog() below), since all but 3 live routes
+// (HEALTH, VERSION, BACKUPS_LIST) are POST; those 3 set method: "GET"
+// explicitly rather than every other entry repeating "POST".
+//
+// params[].bodyField: the real JSON body field name routes.js actually
+// reads for this param (e.g. `body.characterName`), when it differs from
+// the param's own Discord-facing `name` (e.g. "character"). Defaults to
+// the param's own `name` when omitted -- most params match 1:1; only
+// PLAYERS_LINK ("character" -> "characterName") and
+// PLAYERS_INVENTORY_SEARCH ("search" -> "query") need an explicit
+// override today (found via upstream PR #171's review; see
+// discordCommandCatalog.test.js's cross-check against routes.js's real
+// body.<field> reads for the mechanism that keeps this from silently
+// drifting again).
+//
+// Three (group, subcommand) pairs intentionally fan out to two backing
+// routes each -- confirmed against the real, live Discord bot
+// (arrakis-control-panel/src/commands.js) that these are genuinely ONE
+// Discord subcommand each, not two: the bot registers a single "storage"/
+// "find"/"inventory" subcommand and picks which adapter route to call at
+// runtime based on a param value (scope=guild, or search being present).
+// Each fanned-out route entry below carries its own `selector` describing
+// which param/value picks it; the entry with `selector: null` is the
+// default used when no other selector in the pair matches. This is NOT a
+// naming collision to rename away -- inventing distinct subcommand names
+// (e.g. "guild-storage") would contradict this file's own design mandate
+// (see the module header) that catalog names must match what Discord users
+// actually see today, not a fresh naming scheme.
+// Exported (not just module-private) so discordCommandCatalog.test.js can
+// cross-check every declared field/bodyField against buildCommandCatalog()'s
+// real output and against routes.js's real body.<field> reads, without
+// re-deriving or hand-retyping this table a second time inside the test
+// (which would be tautological -- see that test file's own comments).
+export const COMMAND_METADATA = Object.freeze({
   [DISCORD_ADAPTER_ROUTES.HEALTH]: {
     group: "server", subcommand: "health",
     description: "Check the console Discord adapter.",
@@ -66,6 +113,7 @@ const COMMAND_METADATA = Object.freeze({
     // routes.js/adapter.js rather than trusting this table's own claims.
     capability: DISCORD_CAPABILITIES.STATUS_READ,
     routeEnforcesCapability: false,
+    method: "GET",
     params: []
   },
   [DISCORD_ADAPTER_ROUTES.STATUS]: {
@@ -135,6 +183,7 @@ const COMMAND_METADATA = Object.freeze({
     // rather than silently assumed.
     capability: DISCORD_CAPABILITIES.BACKUPS_READ,
     routeEnforcesCapability: false,
+    method: "GET",
     params: []
   },
   [DISCORD_ADAPTER_ROUTES.ANNOUNCEMENTS]: {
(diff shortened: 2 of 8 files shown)


Red-Blink:
Thanks for the thorough PR and testing. I found three contract issues that need to be resolved before this can merge:

1. The generated `player` group contains duplicate `inventory`, `storage`, and `find` subcommands. Please collapse the route variants or give them unique command identifiers, and add a `(group, subcommand)` uniqueness test.

2. Catalog entries do not include HTTP methods or request-field mappings. This leaves the bot needing the same hardcoded dispatch knowledge the catalog is meant to replace. In particular:
   - `character` maps to `characterName`
   - `search` maps to `query`

3. `diagnosticCapability` is defined for `status` but omitted from the generated response, so consumers cannot determine that diagnostic status requires `LOGS_READ`.

CI is green, but these issues need to be resolved for the catalog to be safely consumable by the planned generator/runtime phases.

yacketrj:
Thanks for the review -- all 3 confirmed and fixed in `3c55a66f`.

1. **Duplicate subcommands**: confirmed exactly as reported. Restructured `buildCommandCatalog()` so a `(group, subcommand)` pair now produces one subcommand object with a `routes[]` array (previously one array element per route, which is what produced the visible duplicates). Verified against the real bot (`arrakis-control-panel/src/commands.js`) that `storage`/`find`/`inventory` are each genuinely one live Discord subcommand that fans out to two backend routes at runtime (a `scope=guild` value match for storage/find, a `search`-present check for inventory) -- so the fix merges them into one entry with a `routes[]` array + a `selector` per route (`{ param, equals }` or `{ param, present }`, with `selector: null` marking the default), rather than inventing new subcommand names that don't match what users actually see. Added 2 regression tests: no `(group, subcommand)` pair is ever split across multiple top-level entries, and exactly the 3 known pairs (not more, not fewer) are the ones that fan out.

2. **Method + field mapping**: added `method` to every catalog entry (defaults to `"POST"`; the 3 real `GET` routes -- `HEALTH`, `VERSION`, `BACKUPS_LIST` -- set it explicitly) and `bodyField` to every param (defaults to the param's own name; overridden for the two mismatches you found: `character` → `characterName` on `PLAYERS_LINK`, `search` → `query` on `PLAYERS_INVENTORY_SEARCH`). Added a test that independently parses `routes.js`'s real source to extract each route's actual `body.<field>` reads and cross-checks every declared `bodyField` against that extraction -- not a second hardcoded copy of the same mapping, so it would catch a real future drift, not just a typo in the test itself.

3. **`diagnosticCapability` dropped**: confirmed root cause -- `buildCommandCatalog()`'s output-mapping block explicitly allowlisted which `COMMAND_METADATA` fields survived into each output entry, silently dropping anything not named there. Fixed via destructure-and-spread: only fields that need renaming (`group`/`subcommand`) or coercion (`requiresWritesEnabled`, `routeEnforcesCapability`, `method`) are named explicitly; everything else (including `diagnosticCapability`, and any future field) passes through automatically. Added a generalized test that iterates every real `COMMAND_METADATA` key per route and asserts it reaches the real JSON output unchanged, plus a specific pin for `diagnosticCapability` itself.

All 3 fixes were mutation-tested (reverted each one individually, confirmed the relevant new test fails with a clear message, restored and confirmed green again) rather than just asserted to work. A Layer 2 audit (Security + QA) against this fix also caught and fixed one additional latent issue in the new test helper itself (an unanchored substring match that could have silently extracted the wrong route's body-field block if `routes.js`'s declaration order ever changed) -- not something you flagged, but found while verifying the fix for #2.

Full backend suite: 1169/1169 passing (1162 baseline + 7 new). `semgrep p/default`: 0 findings. `gitleaks`: 0 leaks in this diff. CI green on `3c55a66f` (all 10 checks).

PR #182, opened 2026-08-20
feat(bases): add Give/Fill/Delete container actions to Storage Inventory

yacketrj:
## Risk classification

**Medium.** Blast radius: one console tab (Base Inventory), Storage-group containers only, guarded by three new narrow RBAC actions. No schema migration, no new network surface, no new credential. This PR's own Layer 3 audit (see below) found and fixed one HIGH data-integrity gap (multi-inventory guard) and one MEDIUM DoS gap (unbounded catalog I/O before a batch-size check) before this branch's first push. Since that audit, an additional data-integrity bug (`volume_override` stored per-stack instead of per-unit) and a real, confirmed `position_index` collision risk were found and fixed during live UI validation on a real deployment (`dune-dev`), and a frontend-only visibility-toggle gap (Delete Selected/Delete All not gated by the Give/Fill/Delete toggle) was found and fixed after that. All are covered below with commit references.

**Update (2026-08-19/20):** after this PR was marked ready, automated code review found 6 more issues, and a Layer 2 audit of that fix found 2 more — see "Layer 2 (Implementation) Eight-Hats Audit — Post-Ready Follow-up" below. All 8 are resolved as of commit `64eea58f`.

## Executive Summary

Adds container management (Give, Give Multiple, Fill, Delete Selected, Delete All) to the Base Inventory tab, scoped to **Storage-group containers only**. Refining/Crafting stay read-only to protect active jobs, unchanged.

Today an operator has no way to add items to a base's storage containers except via a disconnected standalone Storage tab that is itself unreachable from the sidebar. This brings that capability into the tab operators already use for base inventory, with the same safety properties the existing single-item delete already has.

## What Changed and Why

- **Ownership re-verified server-side for every mutation**, not trusted from the client — Delete (`deleteMultipleBaseContainerItems`/`deleteAllBaseContainerItems`) and, as of the post-ready follow-up below, Give/Fill/Give-Multiple (`giveItemToBaseContainer`/`fillItemToBaseContainer`/`giveMultipleItemsToBaseContainer`) all share one `resolveOwnedStorageContainer()` helper: the same claim-CTE ownership verification, atomic row lock, and multi-inventory-ambiguity guard. Give/Fill originally used a separate, unscoped, actor_id-only lookup instead (no group filter, no lock shared with the ownership check) — that was a real gap, closed post-ready; see below.
- **New narrow RBAC actions**: `bases:give-item`, `bases:fill-item`, `bases:bulk-delete-items` — following the existing `bases:delete-item` precedent (deliberately split from `bases:mutate`) so no operator's hand-authored policy is silently widened by this change.
- **Give and Fill restricted to `raw_resource`/`refined_resource`/`component` item groups only** (`resolveFillableCatalogItem`), matching each other — container Give/Give Multiple is not meant to hand out weapons, clothing, or schematics.
- **`raw_resource` added to `FILLABLE_GROUPS`**; 19 items tagged with real, individually-verified volumes (previously only `refined_resource`/`component` were fillable at all). `resolveFillableCatalogItem` now also rejects any fillable-group item missing real, positive volume data (post-ready follow-up), closing a latent bypass of the volume cap.
- **Give/Give Multiple insert at the highest unused slot below `max_item_count`** (`nextHighPositionIndex()`) to mitigate a real, confirmed `position_index` collision risk: the live game engine only reads container rows from Postgres at server startup, so a console insert and a live in-game move/pickup can target the same slot while the map keeps running, and the loser is permanently orphaned on the next restart. In-game additions typically fill low-to-high, so inserting from the high end reduces (does not eliminate) the collision window. Fill cannot use the same mitigation — it tops up toward capacity in the same low-to-high direction the engine already fills — and ships with an in-UI warning instead. Full writeup: `docs/incidents/INC-2026-08-19-GIVE-FILL-POSITION-INDEX-COLLISION.md`.
- **`volume_override` is now stored per-unit, not per-stack total.** The live game engine treats a non-null `volume_override` as a per-unit value and multiplies by `stack_size` itself for display; storing the stack total caused every given/filled item to display a wildly inflated in-game volume (confirmed via exact arithmetic match against a real affected row). Every read-side sum (`baseInventory` ×2, `baseContainerListStorage`, `baseContainerSlots`) now multiplies `volume_override * stack_size` to match. Includes a one-time repair script (`console/api/scripts/repair-volume-override.mjs`, dry-run by default, `--apply` to write) for rows already affected before this fix. Full writeup: `docs/incidents/INC-2026-08-19-VOLUME-OVERRIDE-DOUBLE-MULTIPLIED.md`.
- **Real-time volume tracking** surfaced on the Base Inventory tab (tab totals, per-container cards, contents overlay) — previously only slot usage was shown.
- **Neither Give nor Fill ever rejects a request purely for exceeding a container's remaining volume.** Both clamp the requested quantity down to whatever actually fits and report `requested`/`given`/`clamped`, instead of failing outright and forcing the operator to guess a smaller number and retry. Slot count is not clamped the same way — a single give/fill always consumes exactly one slot, so "no slots left" remains a hard rejection.
- **Give and Fill share one item picker and quantity field**, switched by a mode toggle, instead of two separately-stacked, visually-identical panels — consolidated after finding the two-panel layout confusing once both modes were restricted to the same item groups.
- **The whole Give/Fill/Delete surface is hidden by default** behind one explicit visibility toggle (per-row bulk-select checkboxes, Delete Selected, Delete All, and the Give/Fill panel itself), requiring an explicit confirm dialog to reveal — the dialog restates the restart-visibility requirement and the position_index-collision risk. Does **not** gate the pre-existing single-item delete strip (per-row trash icon, selected-slot Remove/Delete button). Turning the toggle off is instant, no confirmation required — hiding a capability is never the risky direction.
- **Deletion no longer requires a verified-stopped map.** The live game engine only reads container rows from Postgres at server startup (never mid-session), so a console delete cannot race a live in-game read the way, e.g., a config file write might. Give/Fill were never gated on map state either, for the same reason (pure inserts).

### Alternate design considered and rejected

Reusing the existing orphaned `removeItemsFromStorage` function for bulk-delete was considered and rejected — it uses the unscoped, actor_id-only lookup (no group filter), which could reach a Refining/Crafting inventory if wired into this UI. Built `resolveOwnedStorageContainer()` + the two new delete functions instead, sharing the same claim-CTE ownership verification the existing single-item delete already has. (Give/Fill briefly repeated the same rejected shape internally before the post-ready follow-up below closed that gap too.)

## Fresh Install / Migration / Break-Fix Behavior

- **Fresh install**: no special behavior — new routes/actions/UI are additive.
- **Migration/upgrade path**: `runtime/data/admin-items.json` is a plain JSON file, no schema migration. New RBAC actions resolve via existing wildcard grants (`bases:*` on admin, `*` on owner) — no policy file needs to change for an existing deployment to retain current admin/owner access. Moderator/player/observer tiers are unaffected (they never had `bases:mutate` or `bases:delete-item` either).
- **Mid-run failure**: every mutation is a single Postgres transaction — a failure mid-batch (e.g. `giveMultipleItemsToBaseContainer` hitting a slot cap on item 3 of 5) rolls back the entire batch, not a partial insert.
- **No credential/key introduced.**

## Real Test Output

Backend (`console/api`), run against a real, isolated PostgreSQL 17 instance (matching this repo's own CI service container), re-verified as of commit `64eea58f`:
```
$ node --test test/*.test.js
# tests 1399
# pass 1399
# fail 0
```
(1391 baseline + 8 new tests from the post-ready follow-up below.)

Frontend (`console/web`): unchanged by the post-ready follow-up (no `console/web` files touched) — last verified at commit `994c7eea`:
```
$ npx vitest run
Test Files  55 passed (55)
     Tests  528 passed (528)

$ npx tsc -b && npx vite build
(clean — 0 errors)
```

New/modified test files: `console/api/test/baseContainerMutationRoutes.test.js` (structural route/guard verification), `baseContainerMutationRoutes.integration.test.js` (real-HTTP integration tests, spawning the actual server against an isolated database; this immediately found and proved a real production bug before merge — see below), `db.test.js`, `baseInventory.test.js`, `policy.test.js`, `rbacParity.test.js`, `adminCatalog.test.js`, `BaseInventoryTab.test.tsx`, `BasesPanel.test.tsx`.

**A real, critical bug was found and fixed by the new integration tests before merge**: `resolveOwnedStorageContainer()` combined `SELECT DISTINCT` with `FOR UPDATE`, which PostgreSQL rejects outright (`FOR UPDATE is not allowed with DISTINCT clause`) — every real call to Delete Selected/Delete All would have 500'd in production. This was invisible to every mocked unit test, since those mocks pattern-match query text and never execute real SQL. Fixed by resolving the `DISTINCT` candidate set in its own CTE first, then joining back to the real inventory row purely to take the lock — `FOR UPDATE` only applies to that final, non-`DISTINCT` join, which Postgres allows.

## Real Security-Check Output

Re-run against the post-ready follow-up commits:
```
$ gitleaks detect --source . --no-git
1 finding, pre-existing and unrelated (docs/integrations/discord-control-bot/admin-guide.md,
untouched by this branch's diff — confirmed via git diff --stat)

$ node --check src/server.js && node --check src/duneDb.js && node --check src/adminCatalog.js
(no output — no syntax errors)
```

`semgrep` could not be run against the post-ready follow-up commits in the reviewing session's sandbox (silent non-zero exit regardless of config — an environment issue in that session, not evaluated one way or the other; disclosed rather than silently omitted). The original `semgrep scan --config auto` output below is from commit `994c7eea` only, before the post-ready follow-up:
```
$ semgrep scan --config auto --error <all 17 changed source/test files>
Findings: 0 (0 blocking)
Rules run: 210
```

## Layer 3 (Integration) Eight-Hats Audit

Per this repo's own audit discipline (Requirement 20), a Layer 3 audit was dispatched as 8 independent Task-tool agents against this feature's full diff, prior to this branch's first push. Full findings register and STRIDE report are posted as an issue comment on the originating issue (#347, in the fork this PR is submitted from). Summary of that audit:

|---|---|---|---|
| Architect | MEDIUM | CSS `nth-child` header rule targeted the wrong column once the `with-checkbox` modifier added a 6th child | **Fixed** — scoped override + structural-alignment test |
| Architect | MEDIUM | `docs/console/base-inventory.md` not updated despite 5 new endpoints on the tab it documents | **Fixed** — full section added |
| Security | MEDIUM | `baseContainerGiveItemsRoute` ran catalog resolution over every item before the 50-item batch cap was checked — event-loop-blocking DoS | **Fixed** — length check moved before per-item processing |
| Security | — | SQL injection, prototype pollution, path traversal, secret redaction, RBAC bypass, confirmation-phrase gating: all verified clean | No finding |
| GRC | CRITICAL | No Layer 1 (design) or Layer 2 (implementation) audit had been dispatched/documented for the originating issue before this Layer 3 pass | **Acknowledged, not retroactively fixable at the time** — closed by the post-ready Layer 2 audit below |
| DBA | HIGH | Same multi-inventory guard finding as Architect/GRC above, independently derived from the schema's documented multi-inventory fact | **Fixed** (same fix) |
| DBA | MEDIUM | Zero real-Postgres integration tests existed for the 3 new mutation functions | **Fixed** — `baseContainerMutationRoutes.integration.test.js` added (see "Real Test Output" above; this suite itself then caught the CRITICAL `DISTINCT`/`FOR UPDATE` bug described above) |
| UI | HIGH | Give/Fill-vs-Delete map-safety asymmetry is real and deliberate but was never explained to the operator — looked like a bug | **Fixed** — explanatory clause added to the map-safety message |
| QA | HIGH | The route-level test suite was 100% structural (source-text regex), zero real HTTP request/response coverage of the 5 new routes | **Fixed** — same integration-test addition as the DBA finding above |
| QA | HIGH | A "batch stops partway through" test was tautological — a static fixture made partial success indistinguishable from total rejection | **Fixed** — mutation-tested, confirmed falsifiable |
| Network / Cloud Security | — | No port/bind/outbound-call/topology/credential/cloud-provider-config change anywhere in the diff | No finding, verified N/A for both hats |

**STRIDE mapping** (full detail in the issue comment): Tampering (multi-inventory silent-wrong-write risk) — fixed, now throws explicitly instead of guessing; Denial of Service (unbounded catalog I/O before the batch-size check) — fixed. No Spoofing, Repudiation, Information Disclosure, or Elevation of Privilege findings were rated above LOW by this pass; the LOW-severity items in that category (thinner bulk-delete audit detail, a wildcard RBAC-action-name collision risk) were tracked as follow-up work and are also resolved in this PR — see "Fixed since the Layer 3 pass" below.

### Fixed since the Layer 3 pass, before this branch's first push

All of the following were follow-up issues filed from the same audit and resolved on this branch prior to submission:

- Bulk-delete/give-multiple held a container's inventory row lock for the full batch duration with no fixed round-trip bound (~4N round-trips for N items). Fixed: a shared helper resolves/verifies the whole batch in a fixed, small number of set-based round-trips (~N+2), with dedicated mutation-tested round-trip-count assertions.
- Bulk-delete's audit trail carried less detail (no position/quality/durability) than the single-item delete's own audit record. Fixed via a shared `auditDetailSelectFragment()` helper bringing both to parity.
- `bases:delete-items` shared a string prefix with `bases:delete-item`, a latent wildcard-policy collision risk (Elevation of Privilege). Renamed to `bases:bulk-delete-items` — zero migration impact, this action was still unreleased when renamed.
- A bulk-delete test's ownership-query mock used unanchored substring matching, currently unique but fragile against a hypothetical future query sharing both substrings. Re-anchored on the query's real, unique column list.
- No frontend test existed for a partial-batch Give Multiple Items failure. Added, proving the backend's real partial-success message reaches the operator through both `onError` and the modal's inline alert.

### Found and fixed after the Layer 3 pass, during live validation (not covered by the audit above)

The following were found through manual live-UI validation on a real deployment (`dune-dev`, combining this feature with an unrelated upstream sync) after the Layer 3 audit above had already been completed and posted, and were not re-audited by a fresh independent 8-hat pass — each is a scoped, independently tested fix, listed here for full disclosure rather than folded silently into the audit summary above:

- **`volume_override` stored per-stack instead of per-unit** (DBA/data-integrity class). The live game engine multiplies a non-null `volume_override` by `stack_size` itself for display; storing the stack total caused every given/filled item to display a wildly inflated in-game volume. Fixed at every write site and every read-side sum; includes a dry-run-by-default repair script for already-affected rows. Full writeup: `docs/incidents/INC-2026-08-19-VOLUME-OVERRIDE-DOUBLE-MULTIPLIED.md`.
- **`position_index` collision risk between a console insert and a live in-game move/pickup** (Tampering-class, STRIDE). Traced end-to-end through the database's own audit log, not inferred. Give/Give Multiple mitigated (insert at the highest unused slot); Fill cannot use the same mitigation by design and ships with an in-UI warning instead. Full writeup: `docs/incidents/INC-2026-08-19-GIVE-FILL-POSITION-INDEX-COLLISION.md`.
- **Give accepted any catalog item, not just raw/refined resources and components** (scope-creep, found via a real clothing item appearing in the Give combobox). Fixed by switching to the same `resolveFillableCatalogItem()` restriction Fill already used, scoped to this Base Inventory tab only.
- **The Give/Fill/Delete visibility toggle did not actually gate Delete Selected/Delete All or their bulk-select checkboxes** — only the Give/Fill panel itself. Turning the toggle off hid Give/Fill but left both destructive bulk-delete actions fully visible and usable on any Storage container with items, defeating the toggle's stated purpose (issue #371). Fixed by introducing a `bulkDeleteAllowed` gate consumed by the bulk-select checkboxes and the Delete Selected/Delete All block; the toggle was relabeled "Give / Fill / Delete Controls" and its confirm dialog updated to state this explicitly. New regression test added; 6 existing tests that had encoded the buggy (ungated) behavior were corrected.
- **A CSS spacing inconsistency** in the restart-visibility warning banner: its bottom margin was self-owned (`12px`) but its apparent top margin was borrowed entirely from an unrelated sibling element's own margin, rather than being a property of the banner itself — coincidentally correct today, but fragile to any future change near that element. Fixed to `margin: 12px 0`, making the spacing symmetric and self-owned.

## Layer 2 (Implementation) Eight-Hats Audit — Post-Ready Follow-up (2026-08-19/20)

This PR was marked ready before the gap the Layer 3 GRC finding above flagged (no Layer 1/2 audit) was actually closed. Automated code review of the ready PR found 6 real issues; closing them the same way every other round in this PR was closed — independently verified, not just fixed and re-asserted.

**Round 1 — automated code review findings** (commit `176bef09`):

| Finding | Severity | STRIDE | Status |
|---|---|---|---|
| Give/Fill/Give-Multiple lacked the multi-inventory ambiguity guard Delete already had | HIGH | Tampering | **Fixed** (issue #381) |
| Give/Fill ownership verified in a separate, unlocked transaction (TOCTOU) from the write | HIGH | Tampering / Elevation of Privilege | **Fixed** (issue #382) |
| `nextHighPositionIndex`'s fallback could return a slot index beyond the container's capacity | MEDIUM | Tampering | **Fixed** (issue #383) |
| `giveMultipleItemsToStorage` did ~3-4 DB round trips/item instead of bulk-delete's own batching fix | LOW (efficiency) | N/A | **Fixed** (issue #384) |
| Volume cap could be silently skipped for a fillable item with missing/zero catalog volume | LOW | Tampering | **Fixed** (issue #385) |
| `baseContainerDeleteSafety`'s group check was dead code at its 3 mutation-route call sites | LOW | N/A | **Documented** as intentional — no behavior change needed (issue #386) |

**Round 2 — Layer 2 audit of the round-1 fix itself** (commit `64eea58f`), dispatched as 8 independent agents per this repo's own governance rule that a single-pass self-review does not count as real eight-hats coverage:

| Hat | Finding | Severity | Status |
|---|---|---|---|
| Architect | None — verified all round-1 fixes correct against the actual code | — | N/A |
| Security | None — independently verified the TOCTOU fix's lock actually serializes against every other mutation path, all queries remain parameterized | — | N/A |
| GRC | `docs/console/base-inventory.md` still referenced the removed `baseContainerOwnedStorageId()` and pre-rename `giveMultipleItemsToStorage()` | **HIGH** | **Fixed** (issue #387) |
| Network | None — confirmed N/A | — | N/A |
| Cloud Security | None — confirmed N/A | — | N/A |
| UI | None — confirmed the frontend never depended on the removed response fields | — | N/A |
| DBA | No test proved the in-memory volume/count accumulator across >1 successful insert in a batch | **MEDIUM** | **Fixed** (issue #388) — mocked + real-Postgres tests added |
| QA | None — independently re-ran the full suite, hand-traced test logic, confirmed mock/real-SQL parity | — | N/A |

Full findings register and STRIDE detail posted as a comment on issue #347. All 8 issues (#381–#388) filed and closed with commit references.

## Reviewer Feedback Tracked Separately

`@drkshrk`'s review also raised design/UX feedback distinct from the automated findings above — filed as its own tracked work rather than folded into this PR silently, since none of it is a defect in what's shipped here:

- #389 — quantity field / item-select box sizing feels cramped
- #390 — design discussion: should Give/Fill eventually cover all valid game items, and should Add/Give's two visually different item-insert surfaces be consolidated
- #391 — add an in-UI tooltip explaining Give vs Fill
- #392 — 0-slot containers open but don't render a Slot Matrix (bug, needs root-causing)
- #393 — feature request: a way to hold a map deliberately stopped for safe inventory editing (explicitly flagged by the reviewer as a pre-existing, upstream-owned design question, not specific to this PR)

None of these block this PR's mergeability as-is; #389/#391/#392 are candidates for a fast-follow if desired.

## What If This Needs To Be Rolled Back

Pure additive change — no destructive schema change. Reverting this commit removes the new routes/actions/UI; the existing Storage-group single-item delete behavior (which predates this PR) is untouched and continues to work.


drkshrk:
## So my thoughts:
- I would potentially swap qty and the select box (size wise) its way to narrow especially when the scroll bar is there looks weird. (maybe irellevant depending on next line)
image

- Seems to only be resources/generic items that can be added through the Give/Fill/Delete toggle. would think it should include all "valid" game items?
  - Would it be easier to re-use the add logic that was already added and build on the ability to "re-fill" and have the toggle just bypass the map down logic so you don't need to restart/stop the map server?
  - Currently with this there is 2 completely different looking add item box's... feels odd because of that.
  - I like the idea -> especially if someone wanted a community chest that re-fills with items at some point.
  - Maybe a tooltip that explains what Give vs Fill does.  I didnt see it cleanly explained anywhere.
- 0 slot containers can be opened now -> but they are not generating the Slot Matrix.  Would expect these to have the appropriate slots shown like all other for consistency?
image


Not directly your PR but this topic due to Red's Change originally:
- Locking inventory editing to the small window of a map restarting can be challenging.  
  - I think we would need to have a way to hold the map in a stopped position easily to edit and then start it when done to accommodate safe editing mo better.
  
## Automated Code Review Scanning had the following comments:

Review complete. 6 findings survived verification out of ~40 candidates from 9 finder angles:

**Top issues (correctness, CONFIRMED):**
1. Give/Fill/Give-Multiple lack the multi-inventory ambiguity guard that Delete has — silently pick the wrong inventory instead of erroring, inconsistent with the exact case this PR fixed for delete.
2. Give/Fill ownership check has a TOCTOU gap — verified in a separate, unlocked transaction from the write, unlike Delete's atomic check-and-lock.
3. `nextHighPositionIndex`'s fallback can return a slot index beyond the container's actual capacity when existing indexes are corrupted.

**Efficiency (CONFIRMED):**
4. `giveMultipleItemsToStorage` wasn't given the round-trip fix the PR applied to bulk-delete — still ~150-300 sequential queries for a 50-item batch under a held lock.

**Lower priority:**
5. Volume cap can be silently skipped for zero-volume catalog items (PLAUSIBLE — not currently triggerable with today's catalog data).
6. A delete pre-check safety guard is dead code (always defaults to "storage") — harmless since real enforcement happens downstream, but misleading.

Two other candidates (duplicate catalog file reads, repair-script overwrite risk) were investigated and refuted by verification.

yacketrj:
Thanks for the review, @drkshrk — addressing both halves of your comment.

## Automated code-review findings — all 6 fixed

| Finding | Severity | Status |
|---|---|---|
| Give/Fill/Give-Multiple lacked the multi-inventory ambiguity guard Delete already had | HIGH | **Fixed** |
| Give/Fill ownership verified in a separate, unlocked transaction (TOCTOU) from the write | HIGH | **Fixed** |
| `nextHighPositionIndex`'s fallback could return a slot index beyond the container's capacity | MEDIUM | **Fixed** |
| `giveMultipleItemsToStorage` did ~3-4 DB round trips/item instead of bulk-delete's own batching fix | LOW | **Fixed** |
| Volume cap could be silently skipped for a fillable item with missing/zero catalog volume | LOW | **Fixed** |
| Delete pre-check safety guard was dead code | LOW | **Documented** as intentional |

Pushed in `176bef09`. I then ran a real Layer 2 (implementation) eight-hats audit against that fix itself — not a single self-review pass — which caught 2 more real gaps (stale docs referencing a removed/renamed function; a real coverage gap where nothing proved the new batch accumulator across more than one successful insert). Both fixed in `64eea58f`. Full findings register + STRIDE mapping for all 8: issue #347 (and mirrored in this PR's own updated body). Upstream CI passed on both pushes.

## Your design/UX feedback

This is real, separate-from-the-code-review feedback, so I filed it as tracked work rather than quietly acting on (or ignoring) it:

- #389 — qty field / item-select box sizing feels cramped, especially with the scrollbar
- #390 — the bigger question: should Give/Fill eventually cover all valid game items, and should Add's existing logic be reused/consolidated with Give's "re-fill without stopping the map" idea instead of two visually different add-item surfaces? Including your community-chest use case.
- #391 — tooltip explaining Give vs Fill
- #392 — 0-slot containers open but don't render a Slot Matrix (this looks like a real bug, not just polish — needs root-causing)
- #393 — your "hold the map stopped for safe editing" point, filed as its own thing per your own framing that it's a pre-existing, upstream design question, not specific to this PR

None of #389/#391/#392/#393 block this PR as-is. #390 is the one worth a real conversation before deciding whether it's in scope for a follow-up here or its own separate design work — happy to discuss.

PR #183, opened 2026-08-20
docs: revise console layered-auth RFC -- post-merge review corrections (follow-up to #170)

yacketrj:
## Purpose

Follow-up to #170. After the merge (with your three review corrections — all three were real gaps our internal audit passes had missed, and the revised §8 now says so explicitly), we ran a further four-lens validation of the merged text: technical-writer, security-architect, UI/UX, and GRC reviewers dispatched independently, with the document's code-fact claims re-verified against the real sources. Verdict: the tier architecture is sound, but an implementer building from the merged text would have to guess load-bearing behavior. This revision closes those gaps. Documentation only, one file, +96/−34.

## The substantive corrections

- **Passkey persistence is now explicitly gated** (§2.2): fresh proof-of-possession is required to *register* a passkey (previously only rotation demanded fresh proof — minting a durable credential is at least as sensitive); the rotation flow shows the passkey list with revoke-all as the default; a standing revoke-all action exists outside the rotation flow; and Tier 3 rotation always requires the current password+TOTP, so a registered passkey can never authorize rotating the credential it would then survive.
- **Total credential loss is answered** (new §3.4): a per-credential loss matrix including the worst case — TOTP device and all recovery codes lost — with the documented host-filesystem reset that re-triggers §4 enrollment, its trust rationale, and the visibility (audit events + startup banner) that makes an unexpected reset loud rather than silent. The first-enrollment race window is acknowledged with its bounds and mitigations.
- **The recovery flow is specified end to end** (§2.3): login-form discoverability, password+code (the code substitutes for the TOTP factor only, never the password), a restricted re-setup session, full-set invalidation on consumption, and passkey visibility at recovery time.
- **An internal contradiction is resolved**: §2.3 said rotation "clears every session"; §5/§6 said credential-type-scoped. Scoped is correct; §2.3 now reads "every password/TOTP-authenticated session."
- **Implementation parameters are pinned** so two implementations produce compatible stores: TOTP RFC 6238 / SHA-1 / 6 digits / 30 s / ±1 step; the recovery-code domain separator, hex+checksum encoding with client-side pre-check and server-side post-limiter re-check; `timingSafeEqual` digest comparison (replacing the "constant-time `Set` lookup" claim — a V8 `Set` does not actually provide that); `runtime/generated/console-second-factor.json` at mode 0600; a 10-minute enrollment-session TTL; and the enrollment redirect/acknowledgment UX states.
- **The Tier 1 fix gains three operational requirements** (§2.1): fail loudly at boot on the half-configured handoff state (which deny-on-empty would otherwise turn into a silent permanent deny), audited denial reason codes that never influence the authorization decision, and a specified, actionable denial message.

## Fresh install / upgrade / break-fix

No change to the design's migration posture — §4 is clarified, not altered. `WEBAUTHN_RP_ID` unset remains byte-identical to today; the mandatory enrollment flow introduced at your review is kept and now fully specified (TTL, redirect behavior, restart semantics, acknowledgment gate). "What happens if it's lost" is now answered for every credential the design introduces, including the honest worst case (§3.4).

## Testing and checks

Documentation-only — no code paths change; §6's test table was extended so every newly specified behavior has a named test home at implementation time. Gitleaks on the commit: no leaks found.

## Review scope, stated honestly

This validation ran four lenses, not our full eight-hat pass; §8 records that explicitly. The eventual Tier 3 implementation still owes the full layered audit before any code PR reaches you.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yacketrj:
Marking this ready for your review. Scope confirmation for triage:

- **One file, docs only** — `docs/rfc-console-auth.md` (+96/−34); no fork-internal artifacts in the diff (verified with `gh pr diff --name-only`).
- gitleaks clean; the doc's §8 issue references are fully-qualified (`yacketrj/…#NNN`) so they don't mis-link to this repo's issue numbers.
- Validation scope is stated honestly in §8 and above: this revision had a four-lens post-merge review (technical-writer, security, UI/UX, GRC), not our full eight-hat pass — the original design (#170) is what went through the three-pass eight-hat Layer 1. No code changes; the design's migration posture is unchanged, only clarified.

Happy to split, trim, or adjust anything if that's easier to review.

PR #189, opened 2026-08-21
fix(autoscaler): rate-limit three ungated log-scanning heal loops

yacketrj:
## Executive summary

`dune-autoscaler` was observed burning 80.99% of a core on one deployment (14.81% on another, same fork). Root cause: three heal-scan functions in `runtime/scripts/autoscaler.sh` were not rate-limited behind the existing `director_heal_due` helper, so each re-ran `docker logs` + a fresh `python3` regex parse on every main-loop tick (`INTERVAL`, default 5s) instead of on a sane interval. This PR adds the same gating most other scans in this file already use, a regression test proving the gate actually works at runtime, and (as of this revision) fixes several real issues a follow-up code review found in the first version of this PR — see "Remediation" below.

## What changed, and why

`scan_proactive_hagga_handoffs`, `scan_deepdesert_loading_responses`, and `scan_named_destination_failures` each shell out to `docker logs` and pipe the output through a `python3` heredoc for regex parsing — expensive per call. Most other heal scans in this file (`scan_igwo_unavailable_maps`, `scan_stale_server_state`, `scan_core_igw_socket_health`, `scan_director_browser_state`, etc.) are gated behind `director_heal_due <key> <interval>`, a small helper that tracks last-run time in a flat state file and short-circuits if the interval hasn't elapsed. These three were not — a straightforward oversight, not a deliberate design choice.

The worst offender, `scan_named_destination_failures`, re-decodes a 10-minute log window across 3 containers roughly 120x more often than needed at the old cadence. Cost scales with how chatty the watched containers' logs are.

**Fix**: add `director_heal_due` gating to all three, using three new overridable env vars, each set to an interval safely shorter than the log window the function reads (so no detection gap opens):
- `DUNE_AUTOSCALER_PROACTIVE_HAGGA_SCAN_SECONDS` (default 15s; log window is `SINCE`, default 30s)
- `DUNE_AUTOSCALER_DEEPDESERT_LOADING_SCAN_SECONDS` (default 15s; same 30s log window)
- `DUNE_AUTOSCALER_NAMED_DESTINATION_SCAN_SECONDS` (default 60s; log window is `NAMED_DESTINATION_SINCE`, default 10m)

Existing event-level dedup (`hub_travel_seen`, `demand_event_seen`, `director_heal_*` state) already prevents duplicate actions for two of the three at the old 5s cadence; the third (`scan_deepdesert_loading_responses`) did not have this dedup and is fixed below.

**Correction to this PR's original claim**: an earlier revision said these three were "the only heal scans in this file not gated" — that was wrong. `scan_travel_demand` also runs the identical pattern (see "Known, deliberately out of scope" below), and `scan_idle_servers`/`scan_reconnect_demand`/`scan_live_player_partition_alignment` are also ungated (lower priority — SQL-bound, not log-parsing-bound). This PR now includes a test (`tests/autoscaler-scan-gating-inventory-test.sh`) that enumerates every `scan_*` function and documents this exact set as a deliberate, explicit exception list, so the claim is now verifiably accurate rather than asserted in prose.

## Remediation (this revision)

A follow-up code review of the first version of this PR found real issues, fixed here:

1. **Real duplicate-publish bug**: `scan_deepdesert_loading_responses` had no dedup guard before its `publish_rmq_json` call (unlike its sibling scans, which all check `*_travel_seen` first) and a dead if/else after it (both branches identical). Since this scan's 30s log window is now wider than its own 15s scan interval, the same log line for one flow could be — and would be — re-detected across 2 consecutive scans, causing a genuine duplicate travel-response publish to the origin game server. Fixed with the same `deepdesert_travel_seen ... && continue` guard its siblings already use, before any side effect.
2. **Silent gate defeat fixed**: none of the three new `*_SCAN_SECONDS` env vars validated numeric input (unlike `DUNE_AUTOSCALER_DEMAND_INTERVAL`, which already does). A malformed override (plausible, since this file's own `SINCE`/`NAMED_DESTINATION_SINCE` use duration-string values like `"30s"`) would silently and completely defeat `director_heal_due`'s gate rather than fail safe — under `set -e`, the arithmetic comparison's "integer expression expected" error doesn't abort inside an `if` condition, so the scan would just run unthrottled every tick again. Fixed: `validate_scan_seconds()` falls back to the default on a non-numeric value, matching the existing `DEMAND_INTERVAL` pattern.
3. **Detection-gap enforcement**: the "no detection gap opens" safety claim was asserted in prose, not validated in code — an operator override that met or exceeded the log window would silently, permanently miss events in the gap (`docker logs --since` is relative to wall-clock invocation time, so a missed window isn't just delayed, it's gone). `validate_scan_seconds()` now clamps the interval below its log window at startup.
4. **CI wiring gap fixed**: `tests/autoscaler-heal-scan-rate-limit-test.sh` (the regression test this PR's QA-hat review credited with closing a CRITICAL coverage gap) was never added to `ci.yml`'s test list, so it never actually ran in CI. Fixed.
5. **Test-quality fixes**: the rate-limit test's function-boundary detection now anchors on the next real function definition instead of a bare `"\n}\n"` (which an embedded Python/awk heredoc could coincidentally contain at column 0); its tail-marker `grep` no longer dies under `set -e` before its own "not found" diagnostic can print; and it now includes a real call-site check — a `docker`-call counter proving `scan_named_destination_failures`'s actual `docker logs` work is suppressed on a second call within the interval, not just the generic `director_heal_due` helper exercised in isolation.
6. **Small efficiency fix**: `scan_named_destination_failures` now calls `docker ps` once per invocation instead of once per source map (3 fewer calls per tick when due).

## Known, deliberately out of scope (not fixed here)

- **`scan_travel_demand`** also runs the identical `docker logs`+`python3`-regex pattern, but from its own dedicated `DEMAND_INTERVAL` (default 2s) loop for real player travel-demand responsiveness — gating it at a multi-second interval like the others risks a real UX regression (delayed server allocation for a travel request) that would likely be worse than its CPU cost. A proper fix would replace its periodic `--since` snapshot with a continuous `docker logs -f` follower (matching the pattern `follow_director_hagga_handoffs` already uses), not a rushed gate. Tracked as a fork-side follow-up rather than included here.
- **`scan_idle_servers`, `scan_reconnect_demand`, `scan_live_player_partition_alignment`**: lower priority (indexed SQL via `psql_value`, not `docker logs`+regex over a large window); also tracked as a follow-up.
- **`director_heal_due`'s timestamp-before-completion semantics**: applying this gate to functions that previously had no persisted marker introduces a (pre-existing, file-wide) crash-recovery tradeoff — if the process crashes between the gate stamping "just ran" and the function's real work completing, a restart suppresses that scan for up to one interval. This is the same tradeoff every other `director_heal_due`-gated scan in this file already accepts; not something this PR's narrower fix should redesign.

## Fresh-install, upgrade, and break/fix behavior

- **Fresh install / upgrade**: no operator action needed. All new env vars are additive with safe, validated defaults.
- **Mid-run failure**: `director_heal_due` only reads/writes a flat TSV state file via an atomic read-filter-write-then-`mv` pattern already used by every other call site in this file; a crash mid-write leaves the previous state intact. See the accepted timestamp-before-completion tradeoff above.
- **No credential/key is introduced by this change.**
- **Real behavior tradeoffs, stated explicitly**: worst-case remediation latency for a stuck "named destination not found" travel failure goes from ≤5s to up to 60s; DeepDesert loading first-detection latency goes from ≤5s to up to 15s. Both are covered by a real-time follower or independent refresh gate in every case except the first detection of a brand-new event, where the latency increase is real but bounded and operator-tunable via the new env vars.

## Testing

```
$ bash -n runtime/scripts/autoscaler.sh
(no output — syntax OK)

$ shellcheck -S warning runtime/scripts/autoscaler.sh
(4 pre-existing SC2034 warnings, all on lines this diff does not touch)

$ shellcheck -S warning tests/autoscaler-heal-scan-rate-limit-test.sh tests/autoscaler-scan-gating-inventory-test.sh
(no output — clean)

$ bash tests/autoscaler-heal-scan-rate-limit-test.sh
autoscaler gates proactive-hagga, deep-desert-loading, and named-destination heal scans behind director_heal_due; director_heal_due itself rate-limits correctly; and scan_named_destination_failures's real docker-logs work is actually suppressed on a second call within the interval

$ bash tests/autoscaler-scan-gating-inventory-test.sh
autoscaler scan_* gating inventory matches the documented allowlist (no new ungated scan slipped in unnoticed)
```

Also manually verified (then reverted): reverting the new dedup guard makes the rate-limit test fail; adding a fake new ungated `scan_*` function makes the inventory test fail — both are real regression checks, not tautological.

## Security check output

No secrets, credentials, network endpoints, or trust boundaries are touched by this diff.

## Eight Hats / STRIDE

Re-run against this revision: no new findings beyond what's disclosed above. On STRIDE specifically — the original revision's table dismissed Denial-of-Service as flatly not applicable; that was too quick, since the root cause being fixed (an autoscaler burning most of a core) is itself a self-inflicted resource-exhaustion condition, and this diff's safety now depends on the validated, clamped interval logic added in the remediation above (previously a silent gap, per item 2/3). With that fixed, DoS risk from this specific diff is low but the category is not simply N/A the way the first revision claimed.

## Known open item — this PR stays in draft

This fix has **not yet been run against a live deployment for a full server session**. This PR stays in draft until that's done.


yacketrj:
Converting back to draft — a follow-up code review (\`/code-review max\`) against this PR found several real issues that need fixing before this is ready:

- A genuine duplicate-publish bug in `scan_deepdesert_loading_responses` (no dedup guard before its side effect, unlike its siblings) that this PR's wider scan interval made more likely to trigger.
- The three new env vars didn't validate numeric input, so a malformed override could silently and completely defeat the rate-limiting gate this PR exists to add.
- The new regression test was never wired into CI, so it never actually ran.
- This PR's own root-cause claim ("these three were the only ungated heal scans") was false — a 4th function (`scan_travel_demand`) is also ungated, though gating it naively risks a real responsiveness regression given its 2s cadence, so that one is being tracked separately rather than rushed.

Fixing all of this on the fork side first (Project-Arrakis/dune-awakening-selfhost-docker#461), then will update this PR's branch once that lands. Apologies for the churn — should have caught these before opening this PR non-draft.

yacketrj:
Rebased onto current \`upstream/main\` (this branch was 126 commits behind after ~4 weeks of upstream releases; now current).

**One new finding while rebasing, in the same spirit as this PR's existing "Remediation" section**: \`scan_rejected_story_returns()\` — a function that didn't exist when this PR was first opened, added later by upstream's own \`adbc6283\` — has exactly the same missing-gate problem this PR exists to fix: it shells out to \`docker logs\` + a Python regex parse on every main-loop tick with no \`director_heal_due\` rate limiting at all. This PR's own \`tests/autoscaler-scan-gating-inventory-test.sh\` (the structural regression guard this PR adds) caught it immediately once merged in — exactly the kind of drift that test is designed to catch. Gated it the same way as the other three, using the existing \`NAMED_DESTINATION_SCAN_SECONDS\` interval (it shares that scan's log window).

Also updated \`tests/autoscaler-heal-scan-rate-limit-test.sh\`'s call-site check: \`scan_named_destination_failures\` was refactored upstream since this PR was opened (\`02f059c6\`, replacing a hardcoded 3-container \`docker ps\` lookup with a DB-driven 5-source query) — the test's mocks were stale against that refactor, silently exercising 0 rows instead of 5. Fixed to match the current implementation.

Re-ran the full local verification after both fixes:

\`\`\`
$ bash tests/autoscaler-heal-scan-rate-limit-test.sh
autoscaler gates proactive-hagga, deep-desert-loading, and named-destination heal scans behind director_heal_due; director_heal_due itself rate-limits correctly; and scan_named_destination_failures's real docker-logs work is actually suppressed on a second call within the interval

$ bash tests/autoscaler-scan-gating-inventory-test.sh
autoscaler scan_* gating inventory matches the documented allowlist (no new ungated scan slipped in unnoticed)

$ bash tests/autoscaler-story-return-test.sh
autoscaler preserves story return state and moves eligible offline pawns
```

All other \`tests/autoscaler-*.sh\` + \`tests/postgres-shutdown-test.sh\` also pass. \`shellcheck -S warning\` on the changed files shows the same pre-existing SC2034 warnings this PR's body already documents, none new.

Still in draft, per this PR's own stated open item — freshness only, not a claim that the live-deployment validation step is done.

PR #192, opened 2026-08-25
Vehicle permissions: transfer to server custodian, and delete

drkshrk:
## Summary
- Adds "Transfer to Server" to the Vehicles → Permissions Owner box, mirroring the existing Bases → Permissions system-custodian transfer (shared roster-mutation engine, same reserved Server/GM identity). Also fixes roster promotion so promoting a new Owner while the custodian holds it removes the custodian instead of leaving it as a stale Co-Owner (applies to both bases and vehicles).
- Adds the ability to delete a vehicle (global Vehicles panel), mirroring Delete Base's safety model: FK-cascade verified against a real production schema dump, same shipped `permission_actor_destroy`/`delete_actors` procedures, its own pending-delete queue for a live map, mandatory pre-delete safety backup, and a `vehicles:delete` IAM action. Adds a vehicle-specific `dune.actor_state` guard (Travel/VehicleBackup/VehicleRecovery), transcribed from the game's own cleanup procedure.
- Closes an RBAC gap found along the way: vehicles had no `POST`/`DELETE` prefix rule, so the new routes needed explicit `REGEX_ACTIONS_BY_METHOD_PATTERN` entries to avoid falling through to a read-only default.

## Test plan
- [x] `console/api` suite: 1541/1541 passing, 0 skipped (real-Postgres integration tests included) via the compliant Docker test runner
- [x] `console/web` suite: 606/606 passing
- [x] Live-clicked through Transfer to Server end to end against a restored production-shaped database (roster preserved, previous owner demoted correctly)
- [x] Live-verified the delete confirm dialog, copy, and backend request wiring; full delete-apply blocked in this local dev harness only by an unrelated pre-existing Windows/CLI limitation, not by this change


drkshrk:
This needs real live testing...  Only automated testing done in test enviornments.

drkshrk:
This needs a good round of live testing.

PR #199, opened 2026-08-27
console: Tier 3 mandatory TOTP second factor with recovery codes (RFC docs/rfc-console-auth.md §2.3/§4), off by default

yacketrj:
## Executive summary

Implements **Tier 3** of `docs/rfc-console-auth.md` — the password tier's mandatory TOTP second factor, with single-use recovery codes for device loss. The RFC itself is already in this repository (Red-Blink#170), so this is the implementation of an accepted design rather than a new proposal.

**It is off by default and stays off until an operator opts in.** `CONSOLE_TOTP_ENABLED` is unset on a fresh install and on every existing install that pulls this, so the login path is byte-identical to today unless it is deliberately turned on. There is no new required environment variable, no migration, and no schema change.

33 files, +6078/−29. The 29 deletions are all in-place edits to existing auth routes; nothing is removed.

## What this changes, and why

Password-only console login means a single leaked or reused password is total compromise of the server. The RFC's answer for the password tier is a mandatory TOTP factor with a recovery path that does not depend on the operator's device.

- **Login** gains a second step when a factor is enrolled: password, then a 6-digit code or a recovery code.
- **Enrollment** is forced on the first password login after the flag is turned on. It runs in a short-lived, non-renewable, enrollment-only session whose tier is not a valid policy tier, so it default-denies everything even if the route allowlist were bypassed.
- **Recovery codes**: 10 single-use codes shown once at enrollment. Using one does not grant a normal session — it grants a re-setup session that forces enrolling a fresh secret, because the authenticator is presumed lost.
- **Credential changes** (password rotation, recovery-code regeneration) require fresh proof of the current credential from the acting session, not just a valid cookie.

### An earlier design that was rejected

The rate-limiter's real-client-IP resolution was first drafted around `CF-Connecting-IP`. That was rejected: most operators of this project do not run Cloudflare at all. The shipped form is a generic, operator-declared trusted-proxy list that works with any reverse proxy setting `X-Forwarded-For`, and that trusts the header **only** when the immediate TCP peer is in that list. With the list empty — the default — behaviour is byte-identical to the current raw socket address.

### Compose wiring

`docker-compose.web.yml` now passes `CONSOLE_TOTP_ENABLED` and `CONSOLE_TRUSTED_PROXY_IPS` into the console container, defaulting to `0` / empty. Without this the flag was unreachable in any real compose deployment — set in `.env`, never seen by the process — which is exactly how it was found: on a live host with the flag already set and a container whose environment did not contain it.

### New dependency

One new runtime dependency: `qrcode` (`^1.5.4`), used to render the enrollment QR. `console/api/package.json` and the lockfile are both updated, so `npm ci` installs it — no manual step.

A regression guard ships with it (`test/declaredDependencies.test.js`): every third-party import under `src/` must be declared in `package.json`. An undeclared import is invisible locally, because a developer's `node_modules` resolves it from whatever was installed at some earlier point, and fails only on a clean `npm ci` — that is, in CI and on every operator's first install. The symptom is also misleading: the server dies at import time, so the failures surface in whichever unrelated integration tests happen to boot a server rather than in the file that added the import. That is exactly how this one was found, and the guard reproduces it.

### Fresh install
`CONSOLE_TOTP_ENABLED` unset. Single-factor password login, exactly as today. No second-factor file is created.
Covered by `enrollmentFlow.integration.test.js` — *"with CONSOLE_TOTP_ENABLED unset, password login is unchanged single-factor"*.

### Upgrade in place
An operator pulling this on a running install sees **no change at all** until they set the flag. Once they do, their next password login does not sign them in — it puts them into a 10-minute enrollment. That is a real behaviour change on an install someone already uses, so it is documented at the point of decision in `.env.example`, in the flag's own comment, with the instruction to read the recovery guide *before* enabling rather than after being locked out.

Turning the flag back off restores single-factor login and leaves enrolled state intact, so re-enabling does not force a second enrollment.
Covered by `upgradePath.integration.test.js` (forced enrollment on upgrade; flag-off rollback preserves state).

### Mid-enrollment failure
The pending secret lives only on the in-memory enrollment session. A crash, restart, or abandoned setup persists nothing — the operator starts again and receives a **regenerated** secret, and the abandoned one cannot confirm. Two concurrent enrollments cannot both win: the second gets 409 and its session is invalidated.
Covered by `enrollmentFlow` (409 race) and `upgradePath` (abandoned enrollment regenerates).

### Broken install — unreadable state
A corrupt second-factor file **fails closed**: 503, no session, no cookie. A corrupt file must never let the password through without the factor.
Covered by `enrollmentFlow.integration.test.js`.

### Broken upgrade — console rolled back below the state version
Distinguished from corruption, and this distinction is the whole point: here the **state is good and the console is behind**. The 503 says to upgrade the console and explicitly does *not* carry the corrupt-file "remove it to re-enroll" remedy, because that remedy would destroy the operator's only second factor. The console does not rewrite or delete the file.
Covered by `enrollmentFlow.integration.test.js` — *"login on a console rolled back below the state's version says upgrade, never delete"*.

### Break glass — what if the operator loses everything?

Stated plainly: **if both the authenticator and every recovery code are lost, there is no login-surface recovery.** That is deliberate — a second factor that could be reset from the login screen would not be a second factor. The way back in is host filesystem access: stop the console, delete `runtime/generated/console-second-factor.json`, restart, and the next password login re-enters normal enrollment. The password is unchanged throughout; this resets the second factor, not the first. Full procedure in `docs/console/two-factor-recovery.md`.

This means anyone with host filesystem access can reset the console's second factor. That is already true of the database, `runtime/secrets/`, and the console binary, so it is not a new exposure — but it is written down rather than left to be discovered.

**This path had a defect, found while writing this PR, and fixing it is the first commit.** Re-enrollment hardcoded its epoch to `0`, while the rollback watermark is a *separate sibling file* that survives deleting the store. The fresh factor therefore landed permanently behind the watermark, and the first use of a recovery code read `epoch 0 < watermark N`, concluded the store was a restored backup, and wiped all ten codes before checking the submitted one. The operator was handed ten recovery codes that were already dead, and would only discover it the next time they had lost their device — precisely when it is their last resort.

Both `enroll()` and `commit()` now seed from the watermark. A fresh install has no watermark and still starts at `0`, so nothing changes for a new operator. The recovery guide's Case 3 was corrected in the same pass: it used to tell operators to delete the watermark too, which was a workaround for this bug and cost them rollback detection permanently.

## What the independent review covered

The multi-agent review (`/code-review ultra`) ran against head `6114e8a2` and returned **zero findings**. Three commits landed after it and were reviewed by hand instead, each recorded on the tracking issue (Project-Arrakis/dune-awakening-selfhost-docker#407) with per-lens evidence:

| Commit | Change | Runtime logic? |
|---|---|---|
| `879e69f6` | declare the `qrcode` dependency + a guard test | no |
| `86d2ad5c` | pass `CONSOLE_TOTP_ENABLED` / `CONSOLE_TRUSTED_PROXY_IPS` through `docker-compose.web.yml` | no (compose `environment:` only) |
| `fc2dab07` | operator upgrade guide + docs index | no |

Stating it this way because "the review passed" and "the review saw the whole branch" are different claims.

## Documentation

- **`docs/console/authentication-upgrade-guide.md`** (new) — the operator walkthrough for an existing install: checklist, update, enable, the enrollment screen step by step, Settings changes, reverse proxies, turning it off, every failure case with its on-screen message, backups, FAQ. Written against the real UI strings and a real in-place upgrade.
- `docs/console/two-factor-recovery.md` (new) — the lockout cases, including the host-side reset.
- `docs/console/API-REFERENCE.md` — four new endpoints; enrollment-scope note.
- `.env.example` — both new variables, with the "read the recovery guide before enabling" warning.
- `docs/README.md` — both console docs indexed.

## Test output

`console/api` — `npm test`:

```
# tests 1761
# suites 1
# pass 1761
# fail 0
# cancelled 0
# skipped 0
# todo 0
```

`console/web` — `tsc -b && vite build`, then `vitest run`:

```
✓ built in 764ms

 Test Files  73 passed (73)
      Tests  686 passed (686)
```

New coverage: `totp.test.js`, `recoveryCodes.test.js`, `secondFactorStore.test.js`, `enrollmentFlow.integration.test.js`, `recoveryLogin.integration.test.js`, `recoveryCodesRegenerate.integration.test.js`, `passwordRotation.integration.test.js`, `secondFactorRestoreDetection.integration.test.js`, `upgradePath.integration.test.js`, `loginRateLimitProxyIp.integration.test.js`, `sessionFixation.test.js`, `SettingsPanel.credentials.test.tsx`, `TotpSetupScreen.test.tsx`.

Every fix in this branch was mutation-tested — the fix is reverted and the suite must fail, on exactly the intended test and nothing else. Two examples:

```
### MUTATION: revert enroll() epoch seeding (the break-glass fix)
not ok 9 - break-glass: after deleting the state file, re-enrolled recovery codes actually work
# pass 8
# fail 1

### MUTATION: make the version error report as generic corruption
not ok 8 - login on a console rolled back below the state's version says upgrade, never delete
# pass 8
# fail 1
```

`rbacParity.test.js` gains a companion test. The two enrollment routes are not IAM-gated (they are confined by the enrollment-scope allowlist instead), and the parity gate's own suggested remedy is "add it to PUBLIC_EXACT". That remedy would silently reclassify a gated route as unauthenticated, so the new test pins that every `/api/auth/2fa/*` path is either enrollment-confined or IAM-gated, and never merely "public".

## Security check output

`semgrep --config p/default` across all 29 changed source files:

```
[semgrep exit: 0]
```

That run started at 1 finding — `detect-non-literal-regexp` on `new RegExp(\`^\\d{${digits}}$\`)` in `auth/totp.js`. Not exploitable, since no caller supplies `digits`, but it was fixed rather than suppressed (a length check plus a hardcoded regex): a suppression would also have masked a future caller that did start passing the value.

`gitleaks detect --log-opts=upstream/main..HEAD`:

```
INF 5 commits scanned.
INF no leaks found
```

`trivy fs --scanners vuln,secret,misconfig --severity HIGH,CRITICAL`:

```
│ api/package-lock.json │    npm     │        0        │    -    │         -         │
│ web/package-lock.json │    npm     │        0        │    -    │         -         │
│ api/Dockerfile        │ dockerfile │        -        │    -    │         0         │
```

## Live verification (in-place upgrade)

Deployed onto a live host running **v1.4.3 with an authenticator already enrolled from an earlier build** — an epoch-less, pre-watermark state file. Sixteen scenarios exercised over HTTP against the running console, in order: flag reaches the container; boot on old-format state; password alone correctly refused; sign-in with the pre-existing authenticator (state migrated in place); replay rejected; password rotation with fresh proof (and back); regenerate codes (watermark created at `0600`); recovery-code sign-in; the re-setup session boxed in with 403 on both a normal route and the regenerate route; QR rendered from the built image; confirm on the new secret; sign-in on the new authenticator; flag-off rollback with state preserved; flag back on with the original authenticator working again. All passed. Full table on Project-Arrakis/dune-awakening-selfhost-docker#407 (Project-Arrakis/dune-awakening-selfhost-docker#407).

That session found the compose-wiring gap above before the first restart, which is the strongest argument for having run it.

## Draft status

Kept as a draft for the maintainer's own review; every gate on the author's side has evidence. One thing not done: the game server itself was never restarted, because none of this touches it.

Flagging one judgement call for the maintainer rather than burying it: `ADMIN_AUTH_DISABLED=1` bypasses the second factor completely, because it bypasses the password too. A boot guard refusing that combination on a non-loopback bind was written and deliberately **excluded** from this PR — it is an independent hardening change, and `ADMIN_AUTH_DISABLED` already opens the console fully today, so Tier 3 does not make it worse. Happy to send it separately if wanted.


yacketrj:
Closing in favour of a Tier-3-only draft from a dedicated branch (`tier3-upstream`). This PR's head branch also carried Tier 1 (Discord OAuth + RBAC); the two halves are now separate stacked drafts so each can be reviewed on its own, and a PR's head branch cannot be changed in place. The replacement is linked from this repository's PR list momentarily; nothing here needs maintainer action.

PR #201, opened 2026-08-28
console: Tier 3 opt-in TOTP second factor with recovery codes (RFC docs/rfc-console-auth.md §2.3/§4), off by default

yacketrj:
## Executive summary

Implements **Tier 3** of `docs/rfc-console-auth.md` — an opt-in, owner-initiated TOTP second factor for the password tier, with single-use recovery codes for device loss. The RFC itself is already in this repository (Red-Blink#170), so this is the implementation of an accepted design rather than a new proposal.

**It is off by default and stays off until an operator opts in.** `CONSOLE_TOTP_ENABLED` is unset on a fresh install and on every existing install that pulls this, so the login path is byte-identical to today unless it is deliberately turned on. There is no new required environment variable, no migration, and no schema change.

**Supersedes #199.** That draft's head branch carried Tier 1 (Discord OAuth + RBAC) as well as Tier 3; the two halves are now separate stacked drafts so each can be reviewed on its own, and since a PR's head branch cannot be changed, #199 is closed in favour of this one. Nothing in this PR references Discord OAuth, the RBAC tier hardening, or the IAM editor — this is the independent base. Head is `tier3-upstream`: the change (`21e8afd3`), a documentation-correction commit (`3f2796d5`), review fixes from a partial review pass (`4b66c280`, `ad486631`), and a completed `/code-review ultra` pass (`5dd9705f`): POST /api/auth/2fa/confirm is now rate-limited per session (it previously allowed unlimited TOTP guessing against a live enrollment session), and cancelling setup now ends the server-side session like every other terminal branch of the flow already did. The Tier 1 draft stacked on it is #202.

41 files, +7303/−34. The 34 deletions are in-place edits to existing auth routes and docs tables; nothing is removed.

## 📋 UAT test plan — [`docs/console/authentication-qa-checklist.md`](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/blob/tier3-upstream/docs/console/authentication-qa-checklist.md)

Parts 1–7 (T01–T25) of this checklist cover this PR — a 25-case manual test plan (real browser, real phone) with a results table. Part 8 (Discord/Tier 1) belongs to the stacked #202, not this PR. **Update:** #202 also added Part 9 (T37–T40) for a Discord OAuth Settings consolidation that layers on top of it -- one of its four cases (T38) exercises this PR's own TOTP-disable route (`/api/auth/2fa/disable`), but only the *new* zero-2FA warning branch that fires when Discord OAuth (Tier 1) is also configured and doesn't cover the acting tier; this route's own Tier 3/4 behavior when Discord OAuth is absent entirely is unchanged and stays covered here, by Parts 1-7. No test case in this PR's own Parts 1-7 needed updating.

## What this changes, and why

Password-only console login means a single leaked or reused password is total compromise of the server. The RFC's answer for the password tier is an optional TOTP factor an operator can turn on, with a recovery path that does not depend on the operator's device.

- **Login** gains a second step when a factor is enrolled: password, then a 6-digit code or a recovery code.
- **Enrollment** is owner-initiated and opt-in (Settings -> Two-Factor Authentication -> Enable), not forced on first login -- corrected 2026-09-02, issue #665, after live-testing feedback that forcing every operator through enrollment with no opt-out was the wrong default. It still runs in a short-lived, non-renewable, enrollment-only session whose tier is not a valid policy tier, so it default-denies everything even if the route allowlist were bypassed.
- **The authenticator's issuer name is the operator's `SERVER_TITLE`** when set (#690), read live from the .env file, not baked into boot-time config -- an operator running several installs sees each labeled by its real server name in their authenticator app instead of several identical "Dune Docker Console" entries. Falls back to the generic app name when unset; no new required config, no migration.
- **Recovery codes**: 10 single-use codes shown once at enrollment. Using one does not grant a normal session — it grants a re-setup session that forces enrolling a fresh secret, because the authenticator is presumed lost.
- **Credential changes** (password rotation, recovery-code regeneration) require fresh proof of the current credential from the acting session, not just a valid cookie. Password rotation revokes every *other* password/TOTP session.
- **Login rate limiter behind a proxy**: an optional `CONSOLE_TRUSTED_PROXY_IPS` list lets the limiter key on the real visitor when the console sits behind a reverse proxy. Unset (default) is byte-identical to today.

### An earlier design that was rejected

The rate-limiter's real-client-IP resolution was first drafted around `CF-Connecting-IP`. That was rejected: most operators of this project do not run Cloudflare at all. The shipped form is a generic, operator-declared trusted-proxy list that works with any reverse proxy setting `X-Forwarded-For`, and that trusts the header **only** when the immediate TCP peer is in that list, reading the rightmost (proxy-appended) address so the client-supplied leftmost value is never trusted.

A second rejected approach, specific to this split: keeping the two tiers in one PR. They share `auth.js` (session scopes), `config.js`, `server.js` and the Settings panel, so one PR was simpler to build — but the maintainer would have had to review an OAuth trust boundary and a second-factor state machine in one diff. Splitting cost a re-derivation of the shared files; the objective gates for that split (this branch's own full suites, and the Tier 1 branch's tree being byte-identical to the previously combined branch) are recorded on Project-Arrakis/dune-awakening-selfhost-docker#575.

### Compose wiring

`docker-compose.web.yml` now passes `CONSOLE_TOTP_ENABLED` and `CONSOLE_TRUSTED_PROXY_IPS` into the console container, defaulting to `0` / empty. Without this the flag was unreachable in any real compose deployment — set in `.env`, never seen by the process — which is exactly how it was found: on a live host with the flag already set and a container whose environment did not contain it.

### New dependency

One new runtime dependency: `qrcode` (`^1.5.4`), used to render the enrollment QR. `console/api/package.json` and the lockfile are both updated, so `npm ci` installs it — no manual step.

A regression guard ships with it (`test/declaredDependencies.test.js`): every third-party import under `src/` must be declared in `package.json`. An undeclared import is invisible locally and fails only on a clean `npm ci` — that is, in CI and on every operator's first install.

### Fresh install
`CONSOLE_TOTP_ENABLED` unset. Single-factor password login, exactly as today. No second-factor file is created.
Covered by `enrollmentFlow.integration.test.js` — *"with CONSOLE_TOTP_ENABLED unset, password login is unchanged single-factor"*.

### Upgrade in place
An operator pulling this on a running install sees **no change at all**, even after setting `CONSOLE_TOTP_ENABLED=1`. The flag only makes "Two-Factor Authentication" available as an option in Settings — a correct password still signs the operator straight in, exactly as before. Enrollment is owner-initiated (Settings → Two-Factor Authentication → Enable), requires fresh password proof, and only then starts the short-lived enrollment session; an operator who never visits Settings and clicks Enable keeps logging in with the password alone indefinitely, flag on or off. (This is a correction from this PR's original design: the RFC's §4 flow below was written for forced-on-login enrollment, which live-testing feedback from the upstream maintainer — "what if I don't want to setup 2FA? Your PRs are forcing the users to do it" — showed was a real adoption blocker for a self-hosted admin tool. Issue #665 reversed it to opt-in; every other mechanic in this PR is unchanged.)

Turning the flag back off restores single-factor login and leaves enrolled state intact (existing TOTP/recovery-code state on disk is never touched by the flag itself), so re-enabling does not force a second enrollment.
Covered by `upgradePath.integration.test.js` — flipping the flag on an existing single-factor install does not, by itself, change how that operator logs in; the owner can opt in from Settings afterward; flag-off rollback preserves enrolled state.

### Mid-enrollment failure
The pending secret lives only on the in-memory enrollment session. A crash, restart, or abandoned setup persists nothing — the operator starts again and receives a **regenerated** secret, and the abandoned one cannot confirm. Two concurrent enrollments cannot both win: the second gets 409 and its session is invalidated.
Covered by `enrollmentFlow` (409 race) and `upgradePath` (abandoned enrollment regenerates).

### Mid-run failure during a credential change
Password rotation writes the new password only after both factors are proven; a failure before that point changes nothing. Recovery-code regeneration invalidates the old sheet and persists the new digests in one store commit; the codes are returned once with `Cache-Control: no-store`. If the response is lost after the commit, the operator regenerates again from Settings — the previous sheet is already dead either way, which the UI says in so many words.

### Broken install — unreadable state
A corrupt second-factor file **fails closed**: 503, no session, no cookie. A corrupt file must never let the password through without the factor.
Covered by `enrollmentFlow.integration.test.js`.

### Broken upgrade — console rolled back below the state version
Distinguished from corruption, and this distinction is the whole point: here the **state is good and the console is behind**. The 503 says to upgrade the console and explicitly does *not* carry the corrupt-file "remove it to re-enroll" remedy, because that remedy would destroy the operator's only second factor. The console does not rewrite or delete the file.
Covered by `enrollmentFlow.integration.test.js` — *"login on a console rolled back below the state's version says upgrade, never delete"*.

### Break glass — what if the TOTP secret or the recovery codes are lost?

Stated plainly: **if both the authenticator and every recovery code are lost, there is no login-surface recovery.** That is deliberate — a second factor that could be reset from the login screen would not be a second factor. The way back in is host filesystem access: stop the console, delete `runtime/generated/console-second-factor.json`, restart, and the next password login re-enters normal enrollment. The password is unchanged throughout; this resets the second factor, not the first. Full procedure in `docs/console/two-factor-recovery.md`.

Losing only the recovery codes (authenticator still works) is not a lockout: regenerate a fresh sheet from Settings with password + current code. Losing only the authenticator: a recovery code signs in and forces re-setup of a new secret.

This means anyone with host filesystem access can reset the console's second factor. That is already true of the database, `runtime/secrets/`, and the console binary, so it is not a new exposure — but it is written down rather than left to be discovered.

A defect in this path was found and fixed before this PR was opened: re-enrollment hardcoded its epoch to `0` while the rollback watermark is a *separate sibling file* that survives deleting the store, so the fresh factor landed permanently behind the watermark and the first recovery-code use wiped all ten codes as a "restored backup" before checking the submitted one. Both `enroll()` and `commit()` now seed from the watermark; a fresh install has no watermark and still starts at `0`. The recovery guide's Case 3 was corrected in the same pass.

## Documentation

- **`docs/console/authentication-upgrade-guide.md`** (new) — the operator walkthrough for an existing install: checklist, update, enable, the enrollment screen step by step, Settings changes, reverse proxies, turning it off, every failure case with its on-screen message, backups, FAQ. Written against the real UI strings and a real in-place upgrade.
- `docs/console/two-factor-recovery.md` (new) — the lockout cases, including the host-side reset.
- `docs/console/authentication-qa-checklist.md` (new) — a 25-case manual test plan (real browser, real phone) with a results table.
- `docs/console/API-REFERENCE.md` — four new endpoints; enrollment-scope note.
- `.env.example` — both new variables, with the "read the recovery guide before enabling" warning.
- `docs/README.md` — the three console docs indexed.

## Test output

`console/api` — `npm ci && npm test` on head `21e8afd3`:

```
# tests 1937
# suites 1
# pass 1937
# fail 0
# cancelled 0
# skipped 0
# duration_ms 114554.609343
```

`console/web` — `npx tsc --noEmit` (exit 0), `npx vite build`, then `npx vitest run`:

```
✓ built in 677ms

 Test Files  77 passed (77)
      Tests  770 passed (770)
```

New coverage: `totp.test.js`, `recoveryCodes.test.js`, `secondFactorStore.test.js`, `enrollmentFlow.integration.test.js`, `recoveryLogin.integration.test.js`, `recoveryCodesRegenerate.integration.test.js`, `passwordRotation.integration.test.js`, `secondFactorRestoreDetection.integration.test.js`, `upgradePath.integration.test.js`, `loginRateLimitProxyIp.integration.test.js`, `sessionFixation.test.js`, `declaredDependencies.test.js`, `SettingsPanel.credentials.test.tsx`, `TotpSetupScreen.test.tsx`.

`rbacParity.test.js` gains a companion test. The two enrollment routes are not IAM-gated (they are confined by the enrollment-scope allowlist instead), and the parity gate's own suggested remedy is "add it to PUBLIC_EXACT". That remedy would silently reclassify a gated route as unauthenticated, so the new test pins that every `/api/auth/2fa/*` path is either enrollment-confined or IAM-gated, and never merely "public".

Every fix in this branch was mutation-tested — the fix is reverted and the suite must fail, on exactly the intended test and nothing else. Two examples, recorded on the tracking issue:

```
### MUTATION: revert enroll() epoch seeding (the break-glass fix)
not ok 9 - break-glass: after deleting the state file, re-enrolled recovery codes actually work
# pass 8
# fail 1

### MUTATION: make the version error report as generic corruption
not ok 8 - login on a console rolled back below the state's version says upgrade, never delete
# pass 8
# fail 1
```

**Update (#690, TOTP issuer):** two follow-up commits on top of the above (`0c61fb8b`, then a same-day correction `54043e76` after the first version's `process.env`-based approach was found live not to reach the container). `console/api` -- `npm test` on head `54043e76`:

```
# tests 2189
# pass 2188
# fail 0
# skipped 1
```
(the one skip is pre-existing and unrelated). New coverage: `config.test.js` (the static-fallback case) and two real, route-level integration tests in `totpOptIn.integration.test.js` -- one confirms a configured `SERVER_TITLE` appears as the real otpauth issuer, the other confirms the fallback when it's unset. `console/web` unaffected (backend-only change).

## Security check output

`gitleaks git --log-opts="main..HEAD"` (v8.30.1):

```
INF 1 commits scanned.
INF scanned ~359077 bytes (359.08 KB) in 377ms
INF no leaks found
```

`semgrep --config p/default --config p/secrets` (v1.174.0) on every changed `.js/.ts/.tsx/.yml/.json`:

```
Scanning 34 files tracked by git with 1077 Code rules:
Ran 250 rules on 34 files: 0 findings.
```

`trivy fs --scanners vuln,secret,misconfig --severity HIGH,CRITICAL` (v0.73.0), repository root, `node_modules` skipped:

```
0 vulnerabilities, 0 secrets in changed files.
2 HIGH misconfigurations, both pre-existing and untouched by this PR:
  DS-0002 orchestrator/Dockerfile  (no USER instruction)
  DS-0029 orchestrator/Dockerfile:6-23  (apt-get without --no-install-recommends)
```

No shell scripts are changed, so `shellcheck` has nothing to run on.

**Update (#690):** `gitleaks git --log-opts="0604cc6c..HEAD"` (the two TOTP-issuer commits): `2 commits scanned... no leaks found`. `semgrep --config p/default --config p/secrets` on the 4 changed files: 1 finding, `unknown-value-with-script-tag` in `config.test.js:11/13` -- pre-existing (present before these commits, confirmed by diffing against the prior head) and a false positive (a self-generated `mkdtempSync` temp-file path in a test fixture, never attacker-controlled); not introduced or worsened by this change, not fixed here.

## Eight-hats + STRIDE audit summary

This code was audited on the author's side under a layered discipline (design → per-feature implementation → full-diff integration), with eight independent review lenses (architecture, security, GRC, network, cloud security, UI/UX, DBA, QA) dispatched as separate workers each round, and every finding mapped to STRIDE. Registers and STRIDE tables are on the tracking issue, Project-Arrakis/dune-awakening-selfhost-docker#407:

| Round | Scope | Outcome |
|---|---|---|
| Layer 1 (design) | RFC §2.3/§4 | findings resolved in the design before code |
| Layer 2 (implementation, per phase) | store / TOTP / enrollment / recovery login / regeneration / session invalidation / frontend | CRITICAL/HIGH resolved per phase, MEDIUM/LOW filed |
| Layer 3 (integration) | eight-hats consolidated register, `/code-review ultra`, then `/code-review max` on the combined branch | 13 fixed (incl. rollback-protection in `commit()`, X-Forwarded-For rightmost parse, one-time-secret `no-store`, null-body guards, rate-limit lockout), 2 deferred with rationale |

STRIDE coverage for this tier, in brief: **Spoofing** — TOTP replay and recovery-code reuse rejected, enrollment session cannot become a normal session; **Tampering** — versioned 0600 state file, rollback watermark, fail-closed on corruption; **Repudiation** — every credential event audited with actor + reason; **Information disclosure** — secrets only in one-time `no-store` responses, hashes at rest, logs redacted; **Denial of service** — separate credential-proof limiter so the login limiter cannot be used to lock out an authenticated operator; **Elevation of privilege** — enrollment-scope allowlist + IAM parity test.

Those audits ran against the combined Tier 1 + Tier 3 branch before the split. The split itself was gated mechanically (see the tracking issue): this branch's full API and web suites, zero references to any Tier 1 module, and the Tier 1 branch's tree byte-identical to the previously audited combined tree.

## Live verification (in-place upgrade)

Deployed onto a live host running **v1.4.3 with an authenticator already enrolled from an earlier build** — an epoch-less, pre-watermark state file. Sixteen scenarios exercised over HTTP against the running console, in order: flag reaches the container; boot on old-format state; password alone correctly refused; sign-in with the pre-existing authenticator (state migrated in place); replay rejected; password rotation with fresh proof (and back); regenerate codes (watermark created at `0600`); recovery-code sign-in; the re-setup session boxed in with 403 on both a normal route and the regenerate route; QR rendered from the built image; confirm on the new secret; sign-in on the new authenticator; flag-off rollback with state preserved; flag back on with the original authenticator working again. All passed. Full table on Project-Arrakis/dune-awakening-selfhost-docker#407.

That session ran on the combined branch. A full server-session run against **this exact head** is the one gate still open below.

## Draft status

Kept as a draft until two remaining gates are met on the author's side: an independent multi-agent review of this re-scoped diff, and a full live server session against this exact SHA. Nothing here touches the game server itself.

Flagging one judgement call for the maintainer rather than burying it: `ADMIN_AUTH_DISABLED=1` bypasses the second factor completely, because it bypasses the password too. A boot guard refusing that combination on a non-loopback bind was written and deliberately **excluded** from this PR — it is an independent hardening change, and `ADMIN_AUTH_DISABLED` already opens the console fully today, so Tier 3 does not make it worse. Happy to send it separately if wanted.


yacketrj:
Thanks for the direct feedback — you're right, and it's fixed.

## What you asked

> what if I don't want to setup 2FA? Your PRs are forcing the users to do it... 2FA should be optional and users can opt in, from the settings page maybe

## What changed

`POST /api/auth/login` no longer redirects into a forced enrollment session when no TOTP factor is configured — a correct password just logs you in, exactly like today, regardless of `CONSOLE_TOTP_ENABLED`. Enrollment is now something the owner starts themselves:

- **`POST /api/auth/2fa/enable`** (new) — from an already-authenticated owner session, re-proves the current password, then mints the same short-lived enroll-scope session the old forced-login path used to mint automatically. `/api/auth/2fa/setup` and `/api/auth/2fa/confirm` are completely unchanged — only the trigger moved.
- **`POST /api/auth/2fa/disable`** (new) — an opt-in with no way back out isn't really optional, so this is new: fresh password + current TOTP code, then wipes the second-factor state outright and revokes any sibling password/TOTP sessions.
- Settings gets a matching "Two-Factor Authentication" section: an **Enable** control when nothing's configured, a **Disable** control (alongside the existing recovery-code regeneration) once it is.
- `CONSOLE_TOTP_ENABLED` now means "make the feature available," not "force it on" — off by default either way, so a fresh install is unaffected until the flag is set, and setting it no longer changes anyone's next login by itself.

`docs/rfc-console-auth.md` §2.3/§4 keeps the original reasoning for why TOTP mattered (single shared credential, no device binding) but now carries a dated correction explaining that the *conclusion* — forcing it — was the wrong trade-off, plus what actually shipped instead. `docs/console/two-factor-recovery.md`, `docs/console/authentication-upgrade-guide.md`, and `.env.example` were all corrected to match (they described the old forced flow as current, which would have been actively misleading).

## Use cases this now covers

- **Don't want 2FA at all**: never touch the Settings control. Password login stays single-factor forever, same as before this PR existed.
- **Want it, but on your own schedule**: enable it whenever from Settings — no surprise mid-login detour.
- **Enabled it, changed your mind**: Disable from Settings, back to password-only immediately.
- **Existing install, flag flipped on later**: login is unaffected; the Settings section just becomes available. Covered by a rewritten upgrade-path test — flipping the flag on an existing single-factor install no longer changes how that operator logs in.
- **Lost device + recovery codes, host reset**: re-enrollment after the documented `console-second-factor.json` reset is now also opt-in from Settings, not automatic on next login — same host-reset procedure, just no longer forces you back into enrollment against your will either.

## Testing / evidence

- Backend: 2053/2054 `node --test` (1 pre-existing unrelated skip) — includes a new dedicated suite (`totpOptIn.integration.test.js`) covering both new routes directly: auth required, wrong-password rejection, rate limiting, 409 on double-enable, 400 on disable-when-unconfigured, and that disable actually revokes sibling sessions while keeping the acting one. Every existing integration test that assumed forced enrollment (`enrollmentFlow`, `upgradePath`, `recoveryCodesRegenerate`, `passwordRotation`, `recoveryLogin`, `secondFactorRestoreDetection`) was updated to opt in explicitly and re-verified green, not just patched to compile.
- Frontend: 786/786 `vitest`, `tsc -b` clean, `vite build` clean.
- `gitleaks`/`semgrep --config auto` clean on the diff.
- CI green on both sides for this exact commit (`6e9d173a`): [this repo's run](https://github.com/Red-Blink/dune-awakening-selfhost-docker/actions/runs/33686323259) and [our fork's run](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/actions/runs/33686324129).

What I have *not* done is a manual end-to-end browser walkthrough as a first-time end user — verification here is the real integration-test suite (spawns an actual console process, real HTTP, real cookies/CSRF, no mocks) plus component-level interaction tests, not a live click-through. Flagging that directly rather than overclaiming, since you specifically asked for the end-user's-shoes perspective — happy to do a live walkthrough and report back if that's still useful before you look further.

Tracked on our side as [#665](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/665).

Red-Blink:
Thanks for the work here. I found a recovery-session issue that needs fixing before merge.

If I open two recovery sessions using different recovery codes, complete authenticator replacement in the first, then return to the second, that older session can still replace the new authenticator. Both confirmations return `200`, and the first replacement authenticator stops working.

Please invalidate any outstanding recovery sessions when the authentication factor changes, and verify the credential generation during confirmation so stale or concurrent requests cannot overwrite a newer authenticator.

Please also add a regression test covering this exact sequence.

yacketrj:
Thanks for the detailed repro — confirmed and fixed.

## What you found
> If I open two recovery sessions using different recovery codes, complete authenticator replacement in the first, then return to the second, that older session can still replace the new authenticator. Both confirmations return 200, and the first replacement authenticator stops working.

## Root cause
Recovery re-setup uses `commit()`, which overwrites the second factor unconditionally — correct for a single deliberate rotation, but with no way to tell "the factor I'm replacing is still the one this session started from" from "a different recovery session already replaced it." The existing rollback-detection `epoch` counter can't be reused for this: it also advances on ordinary recovery-code consumption, so two sessions opened close together both see it move before either has actually changed the factor.

## What changed
- `secondFactorStore.js` gains a new `factorVersion` counter, distinct from `epoch`: it advances ONLY when the TOTP secret itself is replaced (`enroll()`/`commit()`), never on recovery-code consumption alone.
- A recovery login now snapshots `factorVersion` onto the resulting resetup session (`expectedFactorVersion`). `commit()` re-checks it at confirm time and refuses (`stale_generation`, 409) if the live factor has moved since — so whichever session confirms *last*, if the factor already changed underneath it, gets refused instead of silently overwriting.
- A successful confirm also invalidates every other outstanding resetup session outright, so returning to a stale one surfaces an immediate "sign in again" rather than reaching this error at the end of a full re-enrollment attempt.

## Regression test
Added the exact reported sequence to `recoveryLogin.integration.test.js`: two recovery sessions from two different codes, setup on both, confirm the first (must succeed), confirm the second (must be refused), then confirm the first authenticator works and the second's secret never went live. Mutation-tested by reverting the fix — reproduces exactly what you described (both confirms return 200) on the prior code.

## Testing / security status
- Backend: full suite green — `console/api` 2204/2205 `node --test` (1 pre-existing, unrelated sandbox-permission failure in an unrelated backup test, present identically with or without this change).
- New unit coverage in `secondFactorStore.test.js` directly on `commit()`'s new generation check: stale rejection, confirming ordinary epoch-only mutations (recovery-code consumption/regeneration) do NOT trip it, a deleted-store edge case, and confirming the check is opt-in so every other existing caller/behavior is unaffected.
- `gitleaks`/`semgrep --config auto` clean on the diff.
- CI green on this commit (`404b4e65`): https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/actions/runs/34064988901

Commit: `404b4e65`

yacketrj:
Thanks — you're right, and it's fixed. Title and description both still described the RFC's original forced-on-login design, not the actual shipped behavior after #665.

Updated:
- **Title**: dropped "mandatory," now "Tier 3 opt-in TOTP second factor."
- **Executive summary / "What this changes"**: both spots calling TOTP "mandatory" now describe it as opt-in/owner-initiated.
- **"Upgrade in place"**: rewritten to state plainly that setting `CONSOLE_TOTP_ENABLED=1` alone changes nothing about login — it only makes the option appear in Settings. Enrollment only starts when the owner explicitly clicks Enable and proves the current password. The old "next login puts them into a 10-minute enrollment" line is gone; that's the #665-era design, not what's implemented. Also corrected the test-coverage note for `upgradePath.integration.test.js` to describe what it actually verifies now (flag-flip alone doesn't change login behavior).

Verified this matches the real test file (`upgradePath.integration.test.js`, which already asserts "TOTP is opt-in: plain login succeeds with no factor configured") and the real server code (`server.js`: `if (config.authDisabled || !config.consoleTotpEnabled) return grantPasswordSession();`) before pushing the description update, not just editing prose.

Red-Blink:
Thanks for the extensive work here. I found two authentication lifecycle issues that need fixing before merge:

1. **Existing sessions remain valid after TOTP enrollment or recovery re-setup**

   Completing TOTP enrollment or recovery re-setup currently invalidates only the temporary enroll/resetup session. Existing password/TOTP sessions remain authenticated and renewable through `readSession()`.

   I reproduced this by retaining the original owner-session cookie, completing enrollment through the new enrollment cookie, and then successfully using the original cookie afterward.

   On successful `/api/auth/2fa/confirm`, please revoke all existing password/TOTP sessions using the existing `invalidatePasswordSessions()` mechanism.

   Please add regression coverage proving that cookies created before enrollment and before authenticator replacement receive `401` afterward.

2. **Recovery does not perform the atomic invalidation described in `docs/rfc-console-auth.md`**

   `consumeRecoveryCode()` currently removes only the submitted code while leaving sibling recovery codes and the old TOTP secret active until re-setup completes.

   I reproduced both behaviors:
   - Another recovery code can create a second resetup session.
   - The old authenticator can still perform a normal login during recovery.

   The first successful password + recovery-code verification should atomically:

   - Invalidate the complete old recovery-code set.
   - Put the existing TOTP enrollment into a persisted recovery-pending state.
   - Reject normal login using the old TOTP.
   - Reject further recovery attempts from the old code set.
   - Only then issue the restricted resetup session.

Please add integration tests covering sibling recovery-code rejection, old-TOTP rejection, and persistence of the recovery-pending state across process restart.

yacketrj:
Thanks for the detailed repro on both — confirmed and fixed.

## 1. Existing sessions remain valid after TOTP enrollment or recovery re-setup

Root cause: `POST /api/auth/2fa/confirm` only ever ended the temporary enroll/resetup session itself — a standing password/TOTP session created before a factor was enrolled (or before a compromised authenticator was replaced via recovery) stayed valid and renewable.

Fixed: `2fa/confirm` now calls the same session-revocation path password rotation already uses (`invalidatePasswordSessions()`) on every successful enrollment and resetup — no exception for an "acting session," since this route is only ever reached through an enroll/resetup-scope session, which is separately invalidated regardless.

**Found and fixed a real collision this introduced**, before it ever shipped: `invalidatePasswordSessions()` only checked for an empty `userId` to identify a password/TOTP session — but enroll/resetup-scope sessions have an empty `userId` too. Calling it unconditionally would have killed a *different*, still-legitimate concurrent enrollment session before it could reach its own `409 already_configured`. Caught immediately by the existing "second enrollment that loses the race" test going red; fixed by also excluding `session.scope`.

New regression test proves cookies from before enrollment/resetup get `401` afterward, for both the enrollment and the recovery-resetup case.

## 2. Recovery does not perform the atomic invalidation `docs/rfc-console-auth.md` promises

Root cause: `consumeRecoveryCode()` removed only the submitted code from the set, leaving every sibling code and the old TOTP secret live until `commit()` (the actual re-enrollment confirm) finished.

Fixed: the first successful consumption now atomically wipes the *entire* recovery-code set and marks the factor `recoveryPending` — persisted in the same JSON write, so it survives a restart. While pending: `verifyTotpToken()` refuses the old authenticator outright (`recovery_pending`, checked before even verifying the code), and `consumeRecoveryCode()` refuses any further code the same way. Cleared only by a successful `commit()`, which always produces a fresh state object with no pending flag.

New regression tests cover exactly what you asked for: a sibling code rejected once one is pending, the old TOTP rejected mid-window (not just after commit), and the pending state surviving a process restart. Several pre-existing tests whose assertions encoded the old "siblings survive" behavior as correct were updated — including one real bug this surfaced: a test asserting `remaining !== 0` proved "normal consumption, not rollback" no longer had that signal once every real consumption wipes to 0; replaced with what actually proves it (`200` + `resetupRequired`, not `reset_detected`).

## Verification

`console/api`: 2211 → 2218 tests (7 new), 1 pre-existing skip, 1 pre-existing environment-only failure unrelated to this diff (a sandbox `/tmp`-permission test, confirmed via a byte-identical diff against the pre-fix commit). `console/web` unaffected (backend-only). gitleaks and semgrep clean on the diff. CI green on `98088217`.

`docs/rfc-console-auth.md`'s §2.3 "Session invalidation" paragraph is corrected with a dated note — it was already written to describe the intended general rule, but the code only enforced it for password rotation until this fix.

Head is now `306e4d3e`.

Red-Blink:
Reviewed the latest head. The previous concerns around stale recovery sessions, session invalidation, concurrent recovery-code use, proxy handling, and malformed secrets have been addressed with regression coverage.

All checks are green. This looks good to me.

yacketrj:
Quick note on the new commits since your last review: synced this branch forward to your `main`'s real `v1.4.16` tip (was `v1.4.12`, 4 releases / 11 commits behind — party-instance scaling, a Landsraad map-start fix, a Vitest security-patch bump, an addon reward runtime, and an orchestrator container-discovery fix, plus the merges through `tier3-upstream`). Clean fast-forward + merge, no conflicts, full suite re-verified green at each stage.

The other new commits (renaming the Discord adapter's `observer` capability tier to `player`, plus a `DISCORD_OBSERVER_ROLE_IDS` → `DISCORD_PLAYER_ROLE_IDS` env var rename with backward-compat fallback) are unrelated cosmetic/naming work on our side, not a response to anything you flagged — capability set is unchanged, still zero write access at that tier.

Nothing here should need a fresh look on your end unless you want one — flagging for visibility since you'd already approved this PR before these landed.

Red-Blink:
Thanks for the extensive work on this. This is not ready to merge into the current `main` yet.

Please rebase the latest head onto the current upstream `main` and resolve the conflicts in:

- `console/api/test/config.test.js`
- `console/api/test/rbacParity.test.js`
- `console/web/src/api/client.test.ts`
- `console/web/src/styles.css`

The existing checks are green, but they ran against the older branch base and do not validate the rebased result.

After rebasing, please rerun the full:

- API test suite
- Web test suite
- Runtime tests
- Security checks
- Dependency checks
- Release validation

The PR description also says the full live server/session test against this exact head is still outstanding.

Because this changes login, TOTP enrollment, recovery codes, and recovery behavior, please complete that exact-head UAT and record the results before requesting another review.

yacketrj:
## Layer 3 (integration/pre-PR) Eight Hats audit — PR #201, 2026-09-24

Dispatched as 4 independent workers covering all 8 hats against the fresh rebase (`b6356819` on `upstream/main`, `8c6884f3`).

**No CRITICAL or HIGH findings from any hat.** This is a mature diff — reviewers found extensive evidence of prior internal review rounds already fixed in place (TOCTOU races, factorVersion/epoch handling, session-scope exclusion, rate-limiting gaps), and this pass's own findings are MEDIUM/LOW refinements, not defects.

### Findings

|---|---|---|---|
| Architect | MEDIUM | Resetup (recovery-code login) destroys all 10 recovery codes at *consumption* time, not at confirm success, but shares enrollment's flat 10-min non-renewable TTL — an abandoned/interrupted resetup attempt loses the whole code set, leaving only a host-filesystem reset as the way back in. Documented and warned about in `two-factor-recovery.md`, but worth reconsidering. | Open — deliberate tradeoff, maintainer decision needed, not blocking |
| Security | MEDIUM | The pre-existing shared `__global__` login rate-limit bucket (32 attempts/15min) now also absorbs TOTP-verification failures, slightly widening the console-wide-lockout DoS surface. Pre-existing infrastructure this PR extends, not a new hole. | Open — follow-up suggested, not blocking |
| UI/UX | MEDIUM | `RecoveryCodesPanel.tsx` has no copy-to-clipboard/download affordance for the 10 one-time recovery codes — a real person must select-and-copy or hand-transcribe under time pressure. | Open — real UX gap, not blocking |
| DBA + Cloud Security (independently, same finding) | LOW | TOTP secret persists via a bespoke store (`secondFactorStore.js`, `runtime/generated/`) rather than the established `runtime/secrets/`/`getOrCreateSecret()` convention used for comparable credentials (session secret, admin password). Functionally safe — same 0600 mode, same backup coverage, arguably more rigorous (atomic+fsync+watermark rollback detection) — but the RFC doesn't state why it diverges. | Open — one-sentence RFC rationale suggested |
| Security | LOW | `/api/auth/2fa/setup` isn't rate-limited itself (only `/confirm` is); small blast radius since it requires a valid enroll/resetup session. | Open, non-blocking |
| Security | LOW | `secureCookies` defaults false unless `NODE_ENV=production` — pre-existing, not this diff, but this PR raises the stakes of a leaked cookie now that it gates a 2FA-protected session. | Open, doc-callout suggested |
| UI/UX | LOW | No copy affordance for the manual-entry TOTP secret either (same category as the recovery-codes gap, lower severity — one short string vs. ten). | Open, non-blocking |
| Architect | LOW | No "start over" path once `totpRequired` is shown (only "lost access" → recovery entry) — low impact since a wrong password already failed before this point. | Open, non-blocking |
| QA | LOW (informational) | `recoveryCodes.test.js`'s atomicity claim depends on `secondFactorStore.test.js`'s separate concurrency tests to be meaningful — confirmed those exist, so no actual gap. | No action needed |

### Explicitly verified clean (no finding)
- CSPRNG/entropy, constant-time comparison, replay prevention, atomic single-use enforcement (real async-mutex serialization, not just a comment claim)
- Recovery/bypass path cannot silently downgrade 2FA; `expectedFactorVersion` closes a concurrent-recovery-session race
- Watermark/epoch rollback detection catches a resurrected-backup replay and poisons the recovery-code set rather than honoring a spent code
- Session fixation closed across every credential-tier transition; enroll/resetup sessions scope-fenced via `ENROLL_ALLOWED`, verified to sit above the central policy gate in route-registration order
- No shell/path/prototype-pollution injection surface in the new auth files
- Secrets never reach `audit()` calls (Requirement 24 compliant)
- Requirement 0 upgrade path verified directly: flag off = byte-identical login path; sessions are in-memory only so a restart (required anyway) can never resurrect a pre-upgrade session shape; `upgradePath.integration.test.js` proves this against a real spawned server, not just by design claim
- No new ports/binds/DNS/outbound calls — TOTP computation and QR rendering are fully offline (grepped and confirmed)
- No new IAM/cloud credentials; `qrcode` is the only new dependency (local-only, npm supply-chain surface for CI's existing scanners to cover)
- Test suite quality: real server spawns, real concurrency races, real end-to-end HTTP flows — many tests explicitly written as regressions for named prior findings (would fail against pre-fix code, not just pass against post-fix code)

### STRIDE (combined across all 4 dispatches)

|---|---|---|---|
| Spoofing | N/A — session/CSRF binding verified correct | — | — |
| Tampering | N/A — atomic writes, watermark rollback detection verified working | — | — |
| Repudiation | N/A — audit trail comprehensive, secrets excluded | — | — |
| Information Disclosure | TOTP secret storage-location/convention inconsistency (LOW, above) | LOW | Open |
| Denial of Service | Shared rate-limit bucket widening (MEDIUM); resetup full-wipe-on-abandon (MEDIUM, DoS-adjacent to the operator's own recovery path) | MEDIUM | Open, both non-blocking |
| Elevation of Privilege | N/A — `expectedFactorVersion`/`recoveryPending` checks verified to prevent stale-session factor overwrite; scope containment verified structurally sound | — | — |

### Verdict
No CRITICAL/HIGH findings — this PR is not blocked from leaving draft on audit grounds. The MEDIUM items are real, worth a maintainer decision, but are pre-existing-infrastructure extensions or documented tradeoffs rather than defects introduced by this diff. Recommend filing the MEDIUM/LOW items above as tracked GitHub issues per Requirement 20 before merge, rather than leaving them only in this comment.

PR #202, opened 2026-08-28
console: Tier 1 Discord OAuth sign-in with role-based tiers, guided setup and IAM editor (RFC §2.1/§2.1.1) — stacked on #201

yacketrj:
## Executive summary

Implements **Tier 1** of `docs/rfc-console-auth.md` (§2.1, with the §2.1.1 amendment included in this diff) — opt-in **Sign in with Discord** for the console, with the console tier resolved from the member's roles in the operator's Discord server, owner derived from Discord **guild ownership** (never from a role), an optional signed bot handoff as the authoritative tier source, an optional Discord-account-2FA gate per tier, an owner-only guided first-run setup from the sign-in page, a Settings section for the role mapping, an owner-only action-centric Access Control (IAM policy) editor, and the signed-in chip with logout. Alongside it: a tightened default RBAC tier model (admin operates and moderates, changes nothing persistent), player moderation split out of the `players:mutate` economy bucket, and action-level API-key denials for credential/destructive operations.

**Update:** this head now also includes the Discord OAuth Settings consolidation (embedding the guided setup wizard into Settings, plus a soft-disable/re-enable/forget lifecycle and a zero-2FA warning guard) — see its own section below for the full writeup. Both were previously separate fork-internal PRs (`Project-Arrakis/dune-awakening-selfhost-docker#681`, `#683`), now folded directly into this branch since neither touches Tier 3/4 concerns independently of Discord OAuth existing at all.

**Stacked on #201 (Tier 3).** Head `tier1-upstream` = #201's `tier3-upstream` + the Tier 1 change (`dbf832b7`), a code-review fold-in (`174f515f`), a merge of #201's documentation corrections, this branch's own documentation corrections (`b4e8d70f`), and now the Discord OAuth Settings consolidation merge (`b5697db9`) plus a QA-checklist follow-up (`91aac3cb`). GitHub cannot base a cross-fork PR on another fork branch, so this PR's base is `main` and its diff currently *includes* #201's commit; once #201 merges this rebases to the Tier 1 commits alone.

**Everything is inert until an operator configures `DISCORD_OAUTH_*`.** A fresh install and an existing install that pulls this and configures nothing has a login path byte-identical to today (or to #201, if that is enabled). No new required environment variable, no migration, no schema change. The only default that changes for *everyone* is the RBAC tier model — and that only affects installs that already hand out non-owner tiers, which requires the Discord sign-in this PR adds (see "Upgrade in place").

## 📋 UAT test plan — [`docs/console/authentication-qa-checklist.md`](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/blob/tier1-upstream/docs/console/authentication-qa-checklist.md)

Part 8 (T26–T36) covers the original Tier 1 sign-in/role-mapping work — eleven cases with expected on-screen text, stacked on #201's Parts 1–7. **Part 9 (T37–T40, added with this update)** covers the Discord OAuth Settings consolidation below: the disable/re-enable cycle, the zero-2FA guard's both branches, the destructive Forget path, and the already-active-summary UX fix. All four are now Pass, live-verified against the current head (2026-09-03) on a real dev deployment — not just the surrounding UI, the actual security property: for T37 and T39, `/api/auth/discord/start` was checked immediately after the disable/forget request returned, deliberately before the restart could possibly have completed, and was already refused. This UAT pass itself caught and led to fixing one more real, separate bug found the same way as the original compose-passthrough gap below: the new SERVER_TITLE-based TOTP authenticator issuer (#690 — a Tier 3/4 feature added to #201 during this same pass, inherited here through the stack, not new Tier 1 scope) also never reached the container for the identical reason, fixed before continuing. Two more features that layer on top of this branch's own head each have their own dedicated plan: [`discord-role-name-chip-uat.md`](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/blob/feat/573-discord-role-name/docs/console/discord-role-name-chip-uat.md) and [`discord-popup-login-uat.md`](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/blob/feat/574-discord-popup-login/docs/console/discord-popup-login-uat.md) — see #651/#652 on the fork.

## What this changes, and why

- **Sign-in.** `identify guilds guilds.members.read` OAuth with PKCE and a `discord_oauth_state` cookie bound across the cross-site callback. The state cookie is `SameSite=None; Secure` — mandatory for a cross-site redirect — so **Discord sign-in requires an HTTPS-terminating front end**; documented, and password sign-in is unaffected over plain HTTP. Silent `prompt=none` re-auth first, with a single interactive retry on `login_required`/`consent_required`/`interaction_required`.
- **Tier resolution.** Owner comes only from `owner: true` on the operator's home guild; roles map to admin/moderator/player via operator-configured role-ID lists, highest wins; a member holding no mapped role is denied. A signed HMAC **handoff** to a companion bot (fail-closed, `timingSafeEqual`, 30 s freshness) is authoritative when configured — and a *half*-configured handoff disables Discord sign-in with a boot warning rather than silently degrading.
- **Separation of duties.** One Discord role maps to exactly one tier; a mapping that lists a role under two tiers is refused at save, and refused at boot for a hand-edited `.env` (Discord sign-in disabled, password unaffected).
- **Discord-account 2FA gate.** Optional per tier (`DISCORD_OAUTH_REQUIRE_MFA_TIERS`, suggested `owner,admin`) using Discord's own `mfa_enabled` — a gate, not a second enrollment. Finalize **refuses** to require it when the operator's own account lacks 2FA, so an operator cannot lock themselves out on the next restart.
- **Guided setup.** *Set up Discord sign-in* on the sign-in page; **the admin password comes first** (only the console owner can connect a server — otherwise anyone who owns *some* Discord server could point this console at it). The setup-mode round-trip mints **no session**: it hands the operator's identity and their *owned* guilds back to the owner session; finalize requires an owned guild, writes the config, and offers a console self-restart.
- **RBAC hardening** (`policy.js`, `actions.js`, `apiKeyScopes.js`). Admin is denied `settings:*`, credential writes (`server:write-credentials`: Funcom token, server IP), `updates:apply/fix/repair`, `backups:restore/import/delete`, `database:mutate/export`, addon install/update, and the player economy; moderator gains individual kick/ban/teleport; observer is a status-only viewer; player is a tight self-service read set. The five auth-config routes are `settings:*` (owner-only), so an admin-tier Discord session cannot rewrite the OAuth trust anchor. API keys can never reach credential or whole-DB-destructive actions even with the namespace's `write` scope — the key deny is the *sole* barrier there because a key session is synthesised as owner, and `apiKeyScopes.test.js` pins that the set is not silently empty.
- **Access Control editor.** Grid + JSON + a local draft "Test" preview; saves the whole tier-keyed store via `PUT`; wildcard-granted rows are locked rather than snapping back; parameterized-route actions (kick/ban/teleport…) get real checkboxes.

### Earlier designs that were rejected

- **Handoff-or-allowlist only** (§2.1 as accepted). Requires the operator to run the companion bot, or maintain a static owner allowlist, for Discord sign-in to grant any tier. Rejected as the *only* path because most operators of this project do not run the bot; §2.1.1 adds console-native role→tier resolution. The handoff stays authoritative when present.
- **Owner as a configurable role.** Rejected: every holder of a mis-mapped role would silently become owner. Owner is derived from guild ownership, which Discord guarantees is exactly one account.
- **Settings-section-only setup.** Rejected: an unconfigured console needs a working first-run path from the sign-in page, and the setup round-trip must not be startable without the admin password.
- **`SameSite=Lax` for the state cookie.** Rejected — the Discord→console redirect is cross-site, so a `Lax` cookie is silently dropped and every sign-in fails "invalid or expired". Pinned by `oauth.test.js` — *"oauth state cookie is always Secure: SameSite=None mandates Secure, independent of ADMIN_SECURE_COOKIES"*.

### Fresh install
No `DISCORD_OAUTH_*`: `/api/auth/discord/start` is 404, the sign-in page shows the password field and a secondary *Set up Discord sign-in* button. Nothing else changes.
Covered by `oauthRoutes.integration.test.js` — *"Discord OAuth start returns 404 when OAuth is not configured"*.

### Upgrade in place
An install that pulls this and configures nothing sees no change. Things an existing operator should know:
- **Tier defaults tightened.** Any install already running an *earlier* build of this fork's Discord sign-in with admin-tier users will find admin can no longer apply updates, restore backups, edit settings, or touch the economy — by design. Owner keeps everything. Pinned by `tierHardening.test.js`. Stock upstream installs have only the owner (password) session and are unaffected.
- **Re-authorization.** The `guilds.members.read` scope is new; anyone who authorized the application under an earlier scope set is asked by Discord to authorize once more (QA checklist T34).
- **Settings UI change (Discord OAuth consolidation).** An install with Discord OAuth already configured sees its old always-editable manual Client ID/Secret/roles form in Settings replaced by the same guided wizard the sign-in page uses. Reconfiguring reaches the identical flow either way; no config is lost or reset by the upgrade itself.

### Mid-run failure
- **During setup:** the setup-mode round-trip mints no session and persists nothing until finalize; abandoning it leaves the console exactly as before. Finalize writes `.env` values through the existing `updateEnvFileValue` path and drops a `discord-setup-pending-restart` marker so the sign-in page says "configured — restart pending" instead of re-offering setup; boot removes the marker once the config is live. If the owner session expires mid-round-trip, the callback says so and the wizard restarts.
- **During sign-in:** any failure (state mismatch, code exchange error, not a member, MFA gate, handoff outage) renders an HTML page with a link back to the sign-in screen and sets **no** session cookie; failures spend a separate OAuth rate-limit bucket so they cannot lock operators out of the password path. Pinned by `oauthRoutes.integration.test.js` (deny paths assert no `asc_session` cookie).
- **Handoff outage:** with a configured handoff the bot's answer is authoritative, so an outage denies *new* Discord sign-ins (existing sessions unaffected) rather than falling back to a stale allowlist. Deliberate, and documented in the RFC's trade-off paragraph.
- **During a Discord OAuth disable/forget (consolidation):** the cutoff is enforced synchronously, in-process, at the moment the request is handled — not only after the console's own self-restart completes. A closed tab or a restart that never finishes still leaves Discord sign-in genuinely refused for the remainder of that process's life; the restart itself is best-effort cleanup (refreshing the persisted config an operator would otherwise see as stale), not the security boundary. Enable is deliberately unable to reverse a real Forget — Forget sets a separate, restart-only-clearable flag, closing a real gap a Layer 3 review found where Enable could otherwise silently undo a Forget mid-restart with zero credential proof.

### Broken config
Half-configured handoff or a role mapped to two tiers → Discord sign-in disabled at boot with a warning naming the exact value; `/start` and the callback refuse with an actionable page; password sign-in unaffected. Covered by `handoff.test.js`, `roleTiers.test.js`, and the `oauthRoutes` "role mapping unsound" cases. The Settings wizard's map step also checks this instantly and locally now (not just server-side), matching what the old manual form used to do before this consolidation replaced it.

### What if a credential is lost?
- **Discord client secret lost or leaked:** reset it in the Discord Developer Portal and put the new value on the console (replace `runtime/secrets/discord-oauth-client-secret.txt`, mode 0600, or use Settings → Discord OAuth → Change application credentials). Documented under *Rotating the Discord client secret* in the upgrade guide. No lockout: the password path is unaffected. Alternatively, **Forget** the configuration entirely and set it up fresh with a rotated secret from Discord's own portal — the deleted secret's only-ever-one-copy-on-disk invariant means nothing lingers after Forget completes (confirmed: never logged, never echoed, never present in any backup this repo's own scripts take).
- **Bot handoff secret lost:** set a new one on both sides (`runtime/secrets/discord-bot-handoff-secret.txt` + the bot). Until then Discord sign-ins that depend on the handoff are denied (fail-closed); password path unaffected.
- **The Discord server owner account is lost or ownership changes:** the console owner follows Discord ownership at the next sign-in; the admin password remains the break-glass owner path throughout. There is no console-side override — that is the point of deriving owner rather than configuring it.
- **Discord sign-in is entirely unavailable and the operator only ever used Discord to sign in (consolidation):** the admin password is always still the way back in — the Discord OAuth Settings section says so explicitly wherever it's shown, and the guided wizard's map step requires an explicit "I know this console's admin password" acknowledgment before it can be turned on in the first place, precisely to prevent an owner walking into this blind.

## Cross-origin redirect chain test

`oauthRoutes.integration.test.js` — *"Discord OAuth sign-in flow works end-to-end through the real server"* — boots the real console process and a fake Discord API, then drives the full chain over HTTP: `GET /api/auth/discord/start` → 302 to Discord with the `discord_oauth_state` cookie set → callback with `code` + `state` and that cookie sent back → 200 HTML return page and a real `asc_session` cookie → `GET /api/auth/me` with that session resolves `owner`. Sibling tests cover the deny path (403 HTML, no session cookie) and the unconfigured path (404). Stated honestly: the harness sends the cookie explicitly on the callback request, so the *browser's* `SameSite` decision is not exercised by it; the attribute contract (`SameSite=None; Secure`, independent of `ADMIN_SECURE_COOKIES`) is pinned separately in `oauth.test.js`, and the real-browser chain is part of the live-session gate below. The consolidation's own disable/enable/forget routes are same-origin JSON POSTs (no redirect chain of their own) and are instead covered by `discordOAuthDisable.integration.test.js`'s real, subprocess-backed integration suite, including three tests that specifically prove the in-process cutoff at each of the three real gate points (plain start, setup-mode start, callback) takes effect with no restart at all.

## Documentation

- `docs/console/authentication-upgrade-guide.md` — the Discord sections: what you need, creating the Discord application, the guided setup step by step, managing access afterwards, what people will see, the HTTPS requirement, rotating the client secret, upgrade notes, plus the Settings consolidation's own correction block (Login Password/Two-Factor Authentication reordering, not a merged section).
- `docs/console/authentication-qa-checklist.md` — Part 8 (T26–T36, original Tier 1) plus **Part 9 (T37–T40, new)** for the disable/enable/forget lifecycle and the zero-2FA guard.
- `docs/console/two-factor-recovery.md` — corrected to describe the real Settings restructuring (Two-Factor Authentication stays its own section, just reordered below Discord OAuth, not merged into a wrapping section).
- `docs/rfc-console-auth.md` — §2.1.1 amendment (console-native role→tier resolution, guided setup, separation of duties), plus a new Tier-1-credential-loss row in the layered table (an OAuth-only owner locked out by a Discord outage).
- `docs/console-iam.md` — the tier model and the Access Control editor; `docs/console/api-keys.md` — action-level key denials and a data-sensitivity note on `maps: read`.
- `docs/integrations/discord-integration/admin-guide.md` — pointer to console sign-in.
- `docs/console/API-REFERENCE.md` — the OAuth and setup routes; `.env.example` and `docker-compose.web.yml` — every `DISCORD_*` variable with the guidance above, including `DISCORD_OAUTH_DISABLED` (UI-managed, not hand-edited).

## Test output

`console/api` — `npm ci && npm test` on head `91aac3cb` (Tier 3 + Tier 1 + Discord OAuth Settings consolidation):

```
# tests 2382
# pass 2381
# fail 0
# skipped 1
```
(the one skip is pre-existing and unrelated to this diff)

`console/web` — `npx tsc --noEmit` (exit 0), `npx vite build`, then `npx vitest run`:

```
✓ built in 767ms

 Test Files  99 passed (99)
      Tests  1006 passed (1006)
```

Tier 1 coverage on top of #201: `oauth.test.js`, `handoff.test.js`, `roleTiers.test.js`, `oauthRoutes.integration.test.js` (real-server e2e incl. guided setup), `tierHardening.test.js`, `apiKeyScopes.test.js` (key deny set), `rbacParity.test.js`, `IamPolicyEditor.contract.test.tsx`, `IamPolicyEditor.grouping.test.ts`. Discord OAuth Settings consolidation coverage: `discordOAuthDisable.integration.test.js` (real subprocess-backed disable/enable/forget lifecycle, the zero-2FA guard's both branches including the `acknowledgeNoOtherFactor:true` bypass, and the three in-process-cutoff gate points), `envFile.test.js` (the write-queue's concurrency guarantee plus the atomic temp-file+rename crash-safety property, reproduced directly), `DiscordSetupWizard.test.tsx`, `SettingsPanel.discordEmbed.test.tsx`, `SettingsPanel.discordLifecycle.test.tsx`, `App.discordMigrationOffer.test.tsx`, `App.discordSetupReturnMarker.test.tsx`.

## Security check output

`gitleaks git --log-opts="417e56b6..HEAD"` (v8.30.1), scoped to the commits added by this update:

```
INF 9 commits scanned.
INF scanned ~178107 bytes (178.11 KB) in 361ms
WRN leaks found: 6
```

All 6 are the rule `discord-client-id` matching the literal test fixture `123456789012345678` in `DiscordSetupWizard.test.tsx`, `envFile.test.js`, and `discordOAuthDisable.integration.test.js` — a fake, snowflake-shaped placeholder ID, not a real Discord credential. Stated rather than suppressed: this exact literal is a repo-wide convention that predates this update (already used at least twice in `oauthRoutes.integration.test.js` on this branch's prior head) and is used throughout dozens of pre-existing tests; these 6 are simply more instances of it in new test files.

`semgrep --config p/default --config p/secrets` (v1.174.0) on every `.js/.ts/.tsx/.yml/.json` changed since the prior head (18 files):

```
Findings: 0 (0 blocking)
Rules run: 247
Targets scanned: 18
```

`trivy fs --scanners vuln,secret,misconfig --severity HIGH,CRITICAL` (v0.73.0), repository root, `node_modules` skipped: 0 new vulnerabilities/secrets introduced by this diff. Same 2 pre-existing `orchestrator/Dockerfile` misconfigurations (DS-0002, DS-0029) as before, untouched by this diff. One additional finding, **unrelated to this diff** — `runtime/public-probe/go.mod`: `golang.org/x/crypto` CVE-2026-56854 (CRITICAL, fix available in 0.55.0) — this diff touches zero files under `runtime/public-probe/`, so this pre-exists on the branch and would show up on a scan of the prior head too; flagged here for visibility, not something this PR introduces or should be expected to fix.

No shell scripts are changed, so `shellcheck` has nothing to run on.

## Eight-hats + STRIDE audit summary

Original Tier 1 work: same layered discipline and the same tracking issue as #201 (Project-Arrakis/dune-awakening-selfhost-docker#407), where the registers and STRIDE tables live:

| Round | Scope | Outcome |
|---|---|---|
| Layer 1 (design) | RFC §2.1 + §2.1.1 | Steam-flow/OAuth binding and fail-closed handoff design fixed before code |
| Layer 2 (implementation) | handoff deny-on-empty + boot validation, exchange fail-closed, role→tier, guided setup, IAM editor | CRITICAL/HIGH resolved per feature |
| Layer 3 (integration) | eight-hats register, `/code-review ultra`, `/code-review max`, a four-hat pass on the v1.4.4 upstream merge, and a second `/code-review max` on this split diff (15 confirmed findings: 6 HIGH fixed in `174f515f` with failing-first tests — non-owner tiers trapped in the setup wizard, `/start` flood evicting other users' OAuth state, a shared global lockout bucket on the callback, the IAM grid wiping an unparseable draft, compose not passing the two secrets, regex metacharacters in policy patterns; 1 docs fix; 8 filed, medium/low) | fixed: state-cookie `Secure`, self-lockout guard on the 2FA requirement, pending-state DoS eviction, IAM save/test/toggle contracts, admin→owner escalation via `setup:write`, key-deny set + tests; 2 deferred with rationale; MEDIUM/LOW filed |

**Discord OAuth Settings consolidation (new, this update):** own tracking issue, Project-Arrakis/dune-awakening-selfhost-docker#676, full findings register and STRIDE table posted there:

| Round | Scope | Outcome |
|---|---|---|
| Layer 1 (design) | soft-disable vs. hard-wipe tradeoffs, the zero-2FA guard's placement, Tier 1↔3/4 migration matrix | 39 findings, all CRITICAL resolved before implementation began |
| Layer 2 (implementation) | 8 hats dispatched in parallel against the real implementation diff | 1 CRITICAL (the zero-2FA guard's own bypass path was completely untested — confirmed by injecting the exact bug and watching the suite stay green), 5 HIGH (the disable/forget cutoff depending on a second best-effort client call that could be lost entirely; a `.env` write that wasn't crash-safe, reproduced directly; the reconfigure-credentials flow losing its own "restart required" state; two documentation-accuracy defects) — all fixed and verified; 7 MEDIUM filed (#684–#688) with deferral justification; 12 LOW filed as a checklist (#689) |
| Layer 3 (integration) | `/code-review high` against the full branch diff | 2 more HIGH found in the Layer 2 fix's own blind spots (the setup-mode OAuth start route missing the same in-process cutoff check; Enable able to silently reverse a Forget mid-restart with zero proof) — both fixed and covered by new regression tests confirmed to catch their exact bugs; several smaller robustness/polish findings fixed inline (session-expiry handling on the zero-2FA guard's own request path, `.env` permission-mode restoration, a dropped client-side role-conflict check, unstyled danger-zone CSS) |

STRIDE for the original Tier 1 work, in brief: **Spoofing** — PKCE + single-use cookie-bound state, HMAC handoff with freshness, owner from guild ownership only; **Tampering** — one-role-one-tier refused at save and boot, auth-config routes owner-only, `.env` writes validated per key; **Repudiation** — every start/callback/finalize/policy-set audited with reason and Discord user id; **Information disclosure** — client secret only in a 0600 secrets file or `.env`, never echoed (`_discordOAuthSecretSaved` is a boolean), only *owned* guilds retained from the setup round-trip; **Denial of service** — separate OAuth rate-limit bucket, bounded pending-state store, the semgrep ReDoS note above (#576); **Elevation of privilege** — tightened default tiers, action-level key denials, admin cannot rewrite the OAuth trust anchor.

STRIDE for the consolidation, in brief (full table on #676): **Elevation of privilege** — the in-process cutoff flag (and its Layer 3 fix at the setup-mode route) closes the path where a disabled/forgotten Discord OAuth config stayed live until an actual restart; **Tampering** — the `.env` atomic write, the concurrency-test gap filed as #687, a narrow Forget/secret-save race filed as #686; **Repudiation** — Forget's pre-wipe audit log, with a narrow staleness gap filed in #689; **Information disclosure** — the `.env` world-readable permission mode filed as #685; **Denial of service / Spoofing** — no findings mapped this layer.

## Draft status

Kept as a draft until, on the author's side: an independent multi-agent review of this re-scoped diff, and a full live server session against this exact SHA exercising the real-browser redirect chain (password → Discord setup → owner via guild ownership → role→tier → silent re-auth → IAM editor → logout). Discord sign-in and the guided setup/restart were previously validated end-to-end on a dev VM (including a stock-upstream → upgrade path) on the combined branch. The Discord OAuth Settings consolidation's own live cycle is now fully done, against the current head (2026-09-03): disable/verify-the-in-process-cutoff-itself/re-enable, Forget, and the zero-2FA guard's both branches (including the `acknowledgeNoOtherFactor:true` bypass) — see the QA checklist's Part 9 results, all four rows now Pass. The original Tier 1 sign-in/role-mapping live pass (Part 8, T26-T36) against this exact, now-larger head is the one thing still outstanding -- deliberately handed to the operator's own QA process during UAT rather than run here, since it needs several real Discord test accounts with different role assignments, one with Discord 2FA deliberately off, and hands-on control of the Discord server's own role membership (reassigning roles between cases, transferring test-server ownership for T36) -- none of which a scripted session can substitute for. Results will be filled into the QA checklist's Part 8 table once that pass runs.


yacketrj:
Second half of the same feedback, fixed here (this PR is stacked on #201, which carries the 2FA opt-in fix — see that PR's comment for those details).

## What you asked

> when I click on set up discord sign-in, it says Enter the admin password above to set up Discord sign-in. cancel... put yourself in the shoes of the end-user, and go through the process again from scratch

## What changed

The pre-login screen's "Set up Discord sign-in" button used to swap in static text — "Enter the admin password above" — and nothing else: the password field it referred to never received focus, so there was no visual cue where "above" actually was, and the message never said what to do after typing the password.

Fixed narrowly:
- Clicking "Set up Discord sign-in" now focuses the admin-password field immediately (a `ref` + `useEffect` keyed on that button's state).
- The instruction now names the concrete next step: *"Type your admin password above, then select **Sign In** to continue setting up Discord sign-in."*

Deliberately scoped to just this dead-end/focus bug, not the larger guided-setup redesign (have-app/need-app branch, hard HTTPS gate) — that's already tracked separately in #641/#643 and is explicitly gated on its own L1 design pass before implementation, so it wasn't bundled in here.

## Testing / evidence

- New test (`App.discordSetupFocus.test.tsx`) exercises the click → focus → reworded-copy sequence directly against the real component (not mocked click handlers).
- Frontend: 825/825 `vitest`, `tsc -b` clean.
- `gitleaks`/`semgrep --config auto` clean on the diff.
- Our fork's CI green for this exact commit (`e602dbd6`): https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/actions/runs/33686803102

## One thing to flag directly, so it doesn't cost you time

**This PR currently shows `UNSTABLE`/red CI here (`web-tests`, `container-lifecycle`) — that failure is real, but it isn't from this fix.** Root cause is a TypeScript error in `IamPolicyEditor.contract.test.tsx` (an unused `@ts-expect-error` directive, unrelated feature, untouched by this diff). Confirmed: `tsc -b --force` (no cache) passes locally on this exact commit, and the identical commit passes both `web-tests` and `container-lifecycle` on our own fork's CI. Only this repo's runner reproduces it, and the immediately prior commit on this branch passed here fine two days ago. I haven't root-caused the actual environment difference yet — filed as [#673](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/673) with the full comparison rather than guess at a fix in a file I don't otherwise have context on. A retry of the failed jobs would be the cheapest next diagnostic step if you're able to trigger one here.

Same disclosure as #201: verification here is the automated suite plus targeted component tests, not a live manual walkthrough — offering to do that live pass if it's still useful to you.

Tracked on our side as [#666](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/666).

Red-Blink:
I found three additional issues that need fixing before merge:

- **Policy upgrade:** Existing policy files containing the previously supported `observer` tier fail validation, causing all policies to fall back to defaults. I reproduced an admin restricted to `server:read` gaining `server:stop` after upgrade. Please migrate the obsolete tier while preserving the other policies.

- **Discord disable:** A callback already waiting on Discord can still issue an owner session after the disable request succeeds. Please recheck the disable/forget state immediately before creating the session and add a test covering an in-flight callback.

- **Access Control:** When the policy request fails, the early return before the `useMemo` hooks triggers **“Rendered fewer hooks than expected”** instead of showing **Retry**. Please keep the hook order consistent.

This PR also inherits the recovery-session issue I reported on `#201`.

CI and the focused tests pass, but these reproduced cases are still missing from the coverage.

yacketrj:
Thanks — all three reproduced and fixed, plus #201's recovery-session fix merged forward since this PR inherits it.

## 1. Policy upgrade (observer tier)
> Existing policy files containing the previously supported observer tier fail validation, causing all policies to fall back to defaults. I reproduced an admin restricted to server:read gaining server:stop after upgrade.

Root cause: `validPolicyStore()` rejects the WHOLE stored file if it contains any unrecognized tier key, and `"observer"` was retired (folded into `"player"`) with no migration path ever written for existing files. `loadPolicies()` now strips just the obsolete tier document before validation (`migrateObsoleteTiers()`), so every OTHER tier's stored policy loads exactly as authored. The migration is also persisted back to disk so it doesn't repeat (or re-warn) on the next boot, and a boot-time notice logs what was migrated.

Our own follow-up adversarial review of this fix found a related gap: the SAVE path (`setPolicies()`, reachable by pasting old exported JSON into the IAM editor) had no equivalent handling and gave a generic, non-actionable error for the same input. Fixed the same day: it now refuses with a specific message naming the obsolete tier, rather than either silently stripping an explicit save or returning an unhelpful generic error.

## 2. Discord disable / in-flight callback
> A callback already waiting on Discord can still issue an owner session after the disable request succeeds.

Root cause: the callback route's entry gate only reflects the disable/forget flag as of the moment a request first arrives — everything after that (rate limiting, handoff checks, and above all the two real network round trips to Discord) can take long enough for a concurrent Disable or Forget to complete underneath it. `handleOAuthCallback()` now rechecks disable/forget state a second time, immediately before minting a session, and refuses if either flipped during this request's own network calls. (The setup-mode branch was reviewed too — it never resolves a tier or mints a session, so it isn't part of this exposure.)

## 3. Access Control hooks-order crash
> When the policy request fails, the early return before the useMemo hooks triggers "Rendered fewer hooks than expected" instead of showing Retry.

Root cause: the "Failed to load" early return sat before six `useMemo` calls declared further down the component. The first render always fell through and called every Hook; the re-render triggered by the failed fetch hit the early return and called fewer Hooks — React detects the mismatch and throws, crashing the one render meant to show Retry. Both early returns now sit after every Hook in the component, so Hook count/order is identical on every render path; nothing about what either branch renders changed.

## Testing / security status
- A regression test for each issue, each mutation-tested (confirmed to fail on the prior code, pass with the fix):
  - Exact-sequence reproduction for the policy migration (admin restricted to `server:read`, leftover `observer` document, must not regain `server:stop` after load; a second test for the `setPolicies()` follow-up).
  - A real integration test against a fake Discord API whose token exchange is deliberately held open, with a separate authenticated session disabling Discord sign-in while the callback is still in flight.
  - A component test that renders `IamPolicyEditor` against a rejecting mock fetch and asserts the Retry UI actually appears instead of the render throwing.
- Full suites green: `console/api` 2404/2406 `node --test` (2 pre-existing, unrelated — 1 sandbox-permission failure in an unrelated backup test, 1 pre-existing skip), `console/web` 1012/1012 `vitest`, `tsc -b --force` clean.
- `gitleaks`/`semgrep --config auto` clean on the diff (one pre-existing, unrelated semgrep finding on `matchAction()`'s already-defended glob-to-regex conversion, untouched by any of this).
- This PR now also carries #201's recovery-session fix, merged forward (`tier3-upstream` → `tier1-upstream`) — you noted this PR inherits that issue; it's resolved here too.
- CI green on the final commit (`09c19009`): https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/actions/runs/34065709101

Commits: `c7be5322` (policy migration), `2773f73d` (Discord disable recheck), `f0f82581` (hooks order), `978e6030` (merge #201's fix forward), `09c19009` (setPolicies follow-up)

Red-Blink:
The functional changes and focused tests look good, but please remove the accidentally committed Vitest cache.
`node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/results.json`

Also add `/node_modules/` to the root `.gitignore` so root-level generated dependencies and caches cannot be committed again.

Finally, update the PR description where it says Observer remains a status-only tier; the final implementation removes that tier and folds it into Player.

Red-Blink:
The implementation and tests look good. Please update the PR title and description so they match the final behavior: TOTP is optional and owner initiated when `CONSOLE_TOTP_ENABLED=1;` enabling the feature flag alone does not force enrollment on the next login. The current title and the “Upgrade in place” section still describe mandatory/forced enrollment.

yacketrj:
Thanks — you're right, and it's fixed. Just flagging: this specific wording lives on #201 (the stacked base PR — #202 here doesn't carry its own "Upgrade in place" section), so I've replied with the full detail there: https://github.com/Red-Blink/dune-awakening-selfhost-docker/pull/201#issuecomment-5576211171

Summary of what changed on #201:
- **Title**: dropped "mandatory" → "Tier 3 opt-in TOTP second factor with recovery codes... off by default".
- **Executive summary / "What this changes"**: two more "mandatory TOTP" references corrected to opt-in/owner-initiated (same drift, different spots — fixed proactively since the ask was to match the behavior everywhere, not just the one section quoted).
- **"Upgrade in place"**: rewritten to state plainly that setting `CONSOLE_TOTP_ENABLED=1` alone changes nothing about login — it only makes the option appear in Settings. Enrollment only starts when the owner explicitly clicks Enable and proves the current password. The old "next login puts them into a 10-minute enrollment" line is gone.

Verified against the real code (`server.js`: `if (config.authDisabled || !config.consoleTotpEnabled) return grantPasswordSession();`) and the already-correct test file (`upgradePath.integration.test.js`, which already asserted opt-in behavior post-#665) before pushing the description edit — the implementation and tests were already right, only the PR's own English description had drifted.

Checked #202 here too for the same stale language, in case it had crept in independently — it hasn't, no changes needed on this PR's own title/body.

Red-Blink:
I reviewed this incrementally on top of the `#201` head. It inherits the two authentication blockers reported there and introduces an additional credential-management issue.

`readInlineOrFile()` gives `DISCORD_OAUTH_CLIENT_SECRET` precedence over `runtime/secrets/discord-oauth-client-secret.txt`, but the Settings rotation endpoint always writes only the file. The Forget flow deletes only that file and omits `DISCORD_OAUTH_CLIENT_SECRET` from `DISCORD_OAUTH_ENV_FIELDS`.

For installations using the documented inline `.env` variable:

- Rotating the client secret in Settings writes a new file, but the old inline secret remains authoritative after restart.
- The UI reports a successful rotation even though the active credential did not change.
- **Forget this configuration entirely** leaves the inline secret stored in `.env`.
- `unlinkSync()` currently catches every error as “already absent,” so permission or filesystem failures can also produce a false-success response.

Please either update or clear the inline `.env` secret during these operations, or reject Settings-managed rotation/forget with an actionable message explaining that the secret is host-managed.

The API must not report success while a higher-precedence secret remains active. File deletion should ignore only `ENOENT` and return an error for other failures.

Please add integration coverage for:

- Rotation with `DISCORD_OAUTH_CLIENT_SECRET` present.
- Restart-time precedence after rotation.
- Forget with an inline secret.
- Secret-file deletion failures.

There is also a generated Vitest cache committed at:

```text
node_modules/.vite/vitest/.../results.json
```

Please remove it and ensure that path remains ignored.

yacketrj:
Thanks — confirmed and fixed. This PR inherits #201's two authentication-lifecycle fixes (see that PR's comment for the full detail; merged forward here at `306e4d3e`), plus the Discord credential-management issue you found is fixed below.

## DISCORD_OAUTH_CLIENT_SECRET precedence vs. Settings-managed rotate/forget

Root cause: `readInlineOrFile()` correctly gives the env var precedence over the file, but `POST /api/setup/save-oauth-secret` (rotate) and `POST /api/settings/discord-oauth/forget` only ever touched the file:
- Rotating with the env var set wrote a new file that was never actually authoritative — 200 reported, nothing changed after restart.
- Forget deleted only the file, leaving an inline secret fully live — the opposite of what "Forget this configuration entirely" promises.
- `unlinkSync()` caught every error as "already absent," so a real deletion failure (permissions, filesystem) could also report false success.

Fixed with the same env-managed contract this codebase already uses for `ADMIN_PASSWORD` (`config.adminPasswordEnvManaged`): both routes now refuse outright with an actionable message when `DISCORD_OAUTH_CLIENT_SECRET` is set — all-or-nothing, no partial forget — and `unlinkSync`'s catch now only swallows `ENOENT`, returning a real 500 on any other failure.

New `config.discordOAuthClientSecretEnvManaged` is exposed via `publicConfig()`. The Settings UI doesn't yet read it to disable the controls proactively (it currently just surfaces the 400/500 as a runtime error) — filed as a explicit, disclosed follow-up (`dune-awakening-selfhost-docker#709`) rather than expanding this fix's scope, since the backend contract is what actually matters for correctness.

Also removed the committed Vitest cache file you flagged (`node_modules/.vite/vitest/.../results.json`, from a stray top-level `node_modules/` neither of the two scoped `.gitignore` rules covered) and closed that gap.

## Test coverage added, exactly as requested

- Rotation refused with the env var present, file completely untouched.
- The refusal surviving a process restart (not a same-process cache artifact).
- Forget refused with an inline secret, `.env` fields untouched, Discord sign-in still fully live (no partial state).
- A real deletion failure reported as an error, not silent success — the secret path is replaced with a directory to force a privilege-independent `EISDIR` (this suite runs as root, so a chmod-based permission trick wouldn't actually block anything). Mutation-verified: reverted the `unlinkSync` fix and confirmed this exact test catches it, then restored the fix.

3 pre-existing tests that used the inline env var purely as setup boilerplate (not testing the precedence scenario) were moved to file-based secrets, matching what they actually exercise.

## Documentation

`docs/console/API-REFERENCE.md` and `docs/console/authentication-upgrade-guide.md` now state the precedence rule and its Settings-side consequence explicitly ("pick one, not both").

## A broader L3 integration pass (Requirement 20, `/code-review high` against the full PR diff)

Run per this project's own audit discipline. Two more real, pre-existing findings worth flagging directly since they touch this PR's own security model, though neither was part of your review and neither is a regression from anything in this diff:

- **`dune-awakening-selfhost-docker#710`** (HIGH): API-key sessions are hardcoded to `tier: "owner"`, and `apiKeyScopes.js`'s deny lists were never reconciled with `policy.js`'s `CROWN_JEWEL_DENY_ACTIONS` — a key with a broad `players: "write"` scope can currently reach `players:give-item`/`reset`/`recover`/etc., which even a human admin-tier session is denied.
- **`dune-awakening-selfhost-docker#711`** (MEDIUM): `loadPolicies()` never re-applies the crown-jewel Deny backstop `setPolicies()` enforces on save — an operator's pre-existing policy file from before crown-jewel protection existed loads as-is at boot, silently reviving old wildcard admin grants.

Everything else from that pass (10 more findings — duplication, a couple of real-but-minor correctness bugs in the new IAM editor/Discord settings UI, two efficiency items) is batched in `dune-awakening-selfhost-docker#712`. None of these block this PR; flagging for visibility since #710/#711 are real gaps in this exact feature area.

## Verification

`console/api`: 2416/2418 (2 new for this PR's own fix, plus #201's 7), 1 pre-existing skip, 1 pre-existing environment-only failure (same sandbox `/tmp` issue noted on #201, unrelated to this diff). `console/web`: 1019/1019, `tsc` clean. gitleaks and semgrep clean. CI green on `0c82485e`, which is the current head.

Red-Blink:
The default player policy currently grants `players:read` and `guilds:read`, but these permissions expose all players and guilds. The code itself notes that own-player and own-guild scoping is still a follow-up.

Please either implement ownership-based filtering in this PR or remove these permissions from the default player tier until that filtering exists. Ordinary players should not receive server-wide player and guild visibility.

yacketrj:
Thanks — confirmed and fixed. Implemented ownership-based filtering rather than removing the grants.

## What changed

`players:read`/`guilds:read` for the `player` IAM tier are now scoped to the caller's own linked character and guild, not the full server roster:

- `listPlayers()`/`listGuilds()` gained `onlyPlayerControllerId`/`onlyGuildId` filter params (same pattern as the existing `q` search filter). A `player`-tier session's Discord identity is resolved to their linked character via the existing `dune.discord_player_links` table (`getLinkedPlayer()` — this is a real, already-shipped mechanism: `/dune player link <character>` sends a private in-game whisper with a one-time verification code, `/dune player verify <code>` completes the link). Their current guild is resolved via a new `getPlayerGuildId()` helper against `dune.guild_members`.
- `GET /api/players`, `/api/players/online`, `/api/players/search`, and `GET /api/guilds` are all scoped for `player`-tier sessions. `owner`/`admin`/`moderator` sessions are unaffected — the scope resolves to `null` for them, leaving every filter a no-op.
- `GET /api/players/:id` and `GET /api/guilds/:id/members` (the by-id detail routes) previously took no session at all — a `player`-tier caller could read anyone's profile or an arbitrary guild's roster by guessing an id, even with the list views scoped. Both now enforce ownership.
- An unlinked `player`-tier session gets a clear message plus the exact commands to link (`/dune player link`, `/dune player verify`) instead of a bare empty list or an opaque 403. A linked player with no guild gets its own distinct "not currently in a guild" message.

## Verification

New unit tests for both filter params and `getPlayerGuildId()`. Full `console/api` suite: 2448/2449 pass, 1 pre-existing environment-only sandbox permission failure (unrelated, confirmed via clean diff against pre-change state). gitleaks and semgrep clean on the diff. CI green on the updated branch.

yacketrj:
Same sync note as #201 (this branch is stacked on it): fast-forwarded to your `main`'s real `v1.4.16` tip (was 4 releases / 11 commits behind), clean merge through `tier3-upstream`, full suite re-verified green at each stage. Unrelated to your still-open player-tier-scoping comment above, which was addressed separately.

Also renamed this repo's own Discord adapter `observer` capability tier to `player` (plus the matching `DISCORD_OBSERVER_ROLE_IDS` → `DISCORD_PLAYER_ROLE_IDS` env var, with the old name kept as a fallback for existing deployments) — cosmetic naming consistency with our own console-side tier fold, not a response to anything you flagged. Capability set is unchanged.

Red-Blink:
This stacked PR is not ready to merge into the current `main`.

Please first update/rebase `#201`, then rebase this PR on top of the updated `#201` and current upstream `main`.

It currently conflicts in core authorization and server files, including:

- `policy.js`
- `server.js`
- API-key and policy tests
- `App.tsx`
- API-key documentation

Please take particular care when resolving the authorization conflicts. Do not choose one side wholesale.

Current `main` includes newer owner-only system-backup actions that must remain protected from non-owner tiers and API keys:

- `backups:download-system`
- `backups:import-system`
- `backups:restore-system`

Please add or preserve regression coverage confirming those restrictions in the rebased result.

The existing checks passed only against the older branch base. After rebasing, please rerun the complete test, security, and release-validation suites.

The PR description also says the full UAT against this exact head is still outstanding, including:

- Discord sign-in
- Guild-owner handling
- Role-to-tier mapping
- Reauthentication
- IAM editor
- Logout

yacketrj:
## Layer 3 (integration/pre-PR) Eight Hats audit — PR #202, 2026-09-24

Dispatched as 2 independent workers (Security+Architect; QA+GRC+DBA+Network+CloudSec) against this PR's own incremental diff on top of #201 (`git diff origin/tier3-upstream tier1-upstream`, 77 files).

### CRITICAL finding — fixed in `22318232`

**Security hat:** the own-record scoping added for the `player` IAM tier (`players:read`/`guilds:read`) in commit `21113e6f` only covered the list routes (`/api/players`, `/api/guilds`) and `playerProfileRoute` (`GET /api/players/:id`). Every by-id **sub-resource** GET route — inventory, vehicles, bases, currency, solaris-coin, factions, intel, specs, **position**, progression, vitals, teleport-destinations, character-recovery, crafting-recipes, research-items, journey, building-unlocks, customizations, plus the events/stats/history unsupported-feature routes (~20 routes total) — dispatched through `dbPlayerRoute`/`dbPlayerUnsupported`/`buildingUnlocksRoute`/`customizationGrantsRoute`, none of which took a session argument at all. `REGEX_ACTIONS`'s generic `"/api/players/"` → `players:read` prefix rule grants the `player` tier access to all of them. **Any authenticated player-tier session could read any other player's live GPS position, full inventory, currency, vehicle/base locations, and more just by walking sequential ids** — a real stalking/griefing/raid-targeting vector in a PvP survival game, and a materially worse leak than the vulnerability the original commit set out to fix.

**Fix:** added a shared `requireOwnPlayerAccess()` helper (identical behavior to `playerProfileRoute`'s own already-correct check, reusing the same `resolveOwnPlayerScope` resolution so the two checks can't drift), threaded `req` through all four affected route functions. Verified via direct code read (not just trusted from the audit report) before fixing. Full suite re-verified: 2686/2689 api (1 known pre-existing sandbox-only failure, unrelated), 1321/1321 web.

**Known gap, explicitly deferred:** no HTTP-level regression test exists for this fix, or for the pre-existing `playerProfileRoute` check it mirrors — confirmed no test file in this repo exercises any player-tier scoping at the real session/HTTP level anywhere, only unit tests of the underlying SQL WHERE-clause filters (`db.test.js`). Building that harness (real Discord OAuth session + real linked-player DB fixture) from scratch is out of proportion to this fix. **Filing as a separate follow-up issue rather than silently leaving untested** — see the linked issue below.

### Other findings (non-blocking)

|---|---|---|---|
| Architect | MEDIUM | Stale-session tier persists past a Discord-side role revocation for up to 12h (renewable) — no periodic re-validation. | Open — accepted-limitation candidate, needs a maintainer decision on whether to add periodic re-check or just document the TTL tradeoff |
| Architect | MEDIUM | `policy.js` comment still says own-record scoping is "a follow-up" — now stale/misleading given the fix above. | Should be updated in the same pass as this comment's fix, to enumerate exactly which routes are scoped |
| QA | HIGH→addressed by the CRITICAL fix above | Missing test coverage for the enforcement path (not the underlying filter functions) | See "Known gap" above |
| GRC | HIGH (process) | #201/#202 were already marked `isDraft:false` on Red-Blink before this L3 audit ran, and before a referenced fork design-doc PR (#677) had merged. | Noting for the operator's awareness — not reverting draft status unilaterally; this comment documents the audit trail that was missing at the time they were marked ready. |
| GRC | LOW | Unverified whether "found by upstream review 2026-09-06/09-09" commit-message framing reflects real upstream thread comments (no network access to cross-check in this pass). | Non-blocking, flag for whoever posts this |
| DBA | LOW | `getPlayerGuildId()`'s `LIMIT 1` assumes one-guild-per-player without a confirmed DB constraint backing it. | Non-blocking, informational |
| DBA/Architect | LOW | `playerProfileRoute`'s own-record check takes only `rows[0]` from a `pageSize:1` query — a controller with >1 matching row (e.g. stale rows from character recreation) could false-403 the real owner. Fails closed, not exploitable. | Non-blocking |

### Verified clean (no finding)
Discord OAuth flow (real end-to-end test with cookies, satisfies Req 19 gate (e)), IAM editor crown-jewel backstop, credential storage (matches established `runtime/secrets/` convention exactly, redacted in audit logs, never sent to the browser), schema/migration (none — all reads, existing tables), network topology (all endpoints operator-configurable, not hardcoded), rate-limiting on OAuth endpoints.

### STRIDE

|---|---|---|---|
| Information Disclosure | The CRITICAL IDOR above | CRITICAL | **Fixed** (`22318232`) |
| Elevation of Privilege | Same finding, EoP-adjacent (reading another player's data via an authenticated-but-unauthorized path) | CRITICAL | **Fixed** |
| Repudiation | PR marked ready without a completed audit trail (GRC-1) | HIGH (process) | Noted, not a code defect |
| Spoofing / Tampering / Denial of Service | No findings | — | — |

PR #206, opened 2026-08-29
Rework the Home screen status layout

drkshrk:
The hero panel was the only `.grid` child without `.wide`, so the right half of row 1 sat empty on every viewport above the 900px breakpoint. It now spans the row, leads with the overall verdict, and uses the reclaimed space for the six readiness subsystems as rows that route to the tab handling each one.

Also in this pass:

- Performance badges and meter bars derive from each reading rather than "did a sample arrive", so a 96%-full disk no longer shows the same green pill as a 4%-full one. A null reading (the API returns one on the first sample after every restart) reports INFO rather than a green OK.
- The Server Identity band is removed; it repeated the hero summary line. Server IP and Battlegroup move into the hero and render even when unknown, since that is now the only place they appear.
- The status dot splits the overloaded amber (six of ten readings) by what the operator should do: hue for severity, a pulse for in-progress, ring style for no reading yet. It takes the worst of the overall reading and all six subsystems, so a failed Funcom/FLS shows red even while the heading reads OK.
- The three polling effects swallowed their errors, leaving stale values on screen indefinitely. The panel now dates the data and warns after consecutive misses, without blanking the last good values.
- `homeOverallBadge` no longer reports a failed readiness check as healthy — `inferStatus` was matching the word "checked" against its pass list.

Everything is rendering-only. `summarizeHomeStatus` keeps its exact shape because `isHomeActionComplete` and `homeNeedsWarmRefresh` read its `health` and `identity` arrays to drive the restart lifecycle; a regression test pins that and fails if the health array is emptied.

The later two commits address findings from a review pass over the first: two layout bugs (the subsystem value refused to shrink and crushed the label to a 5px column; an unspaced server Title overflowed its track and was painted under the opaque subsystem list), plus accessibility and signal fixes — a visually hidden h2 restoring the heading level and the region's accessible name, a dashed ring so "checking" and "no reading" stay distinct under prefers-reduced-motion, and the population warning regaining a renderer.

Verified: `tsc --noEmit` clean, 784/784 tests across 76 files with 0 skipped. Layout and colour choices were measured in the browser rather than eyeballed.

### DIFF
--- a/console/api/src/config.js
+++ b/console/api/src/config.js
@@ -2,6 +2,7 @@ import { existsSync, mkdirSync, readFileSync, writeFileSync, chmodSync, statSync
 import { dirname, resolve } from "node:path";
 import { createHash, randomBytes } from "node:crypto";
 import { networkInterfaces } from "node:os";
+import { clampInt } from "./jsonStore.js";
 
 export const APP_NAME = "Dune Docker Console";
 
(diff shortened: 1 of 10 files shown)


drkshrk:
still playing with stuff closing this for now.  thought i was closer to having something ready

PR #208, opened 2026-08-31
Surface encrypted system backups in the console, with restore

drkshrk:
`dune db backup-system` already produced an encrypted archive of the database plus `.env`, `runtime/generated/` and `runtime/secrets/`, but it was CLI-only with no restore. This surfaces it in the console and closes the round trip, so a server can be moved to new hardware from a browser.

### Ships

- **Create, list, download, delete.** Download serves the archive and its `.yaml` sidecar as a single streamed `.tar`, and **import** takes that same file back.
- **Restore** — a `dune db restore-system` verb plus a preview-then-apply console flow. Apply is refused until a dry run has succeeded, and that rule is enforced by the API rather than the page (see below).
- **Restore from first-run setup.** The wizard offers it beside "deploy a new server" and runs install → dry run → apply → start → console reload as a checklist. It can be retried after a failure, and resumes if the page is reloaded mid-restore.
- **What a fresh host needs to receive one**: Postgres starts itself, `dune update install-assets` installs game files without touching the database, and the console restarts itself afterwards to load the restored `.env`.
- **The install step says what it is doing.** SteamCMD reported nothing for the whole depot fetch — minutes of an apparently dead step. `install-assets` now names each phase, and on completion the row reads `build 2064155-0-shipping · 11 images · 4.9 GB`, each part dropped when the install did not report it.
- The console image moves to `node:20-trixie`. `--aead-algo` needs gnupg ≥ 2.3 and the image shipped 2.2.40, so `backup-system` had never worked from the console. CI now builds this image and checks that the gpg it ships carries the flag.

### Behaviour changes for existing deployments

Three, all consequences of the security work below. None affects the `owner` tier or the console UI.

- **The `admin` tier no longer holds `backups:download-system`, `backups:import-system` or `backups:restore-system`.** It keeps read, create, `create-system` and `delete-system`.
- **An API key stored as `{"backups": "write"}` no longer reaches those same three.** They are still grantable, but only by naming them in the key's explicit action list.
- **`POST /api/backups/system/{name}/restore` returns 400 for an `apply` value it does not recognise**, rather than silently treating it as a dry run and returning 202. `true`/`false`, `1`/`0`, `""` and omitting the field all behave exactly as before.

### Security fixes found reviewing this branch

- **Preview-then-apply was enforced only by the React flow**, reported by @Red-Blink. A caller holding the restore grant could POST `apply: true` and replace `.env`, `runtime/secrets/`, `runtime/generated/` and the database without ever previewing — and `runner.js` sets `DUNE_DB_ASSUME_YES`, so `db.sh`'s typed confirmation did not gate that path either. A successful dry run is now recorded against the calling session or API key, the archive, and a sha256 taken before the run; an apply without a matching, unexpired record is refused with 409.
- **The approved archive is pinned to its digest.** The console hashes in the apply request, but the restore spawns asynchronously and `db.sh` opens the file seconds later — and the import route can rename a different archive onto that name in between. `restore_system` now copies the archive into its own private staging directory, hashes that copy, and decrypts the same copy, so the bytes checked and the bytes used are one file nothing outside the process can reach.
- **Adding the five system-backup actions was an `admin` → `owner` escalation.** The archive holds `runtime/secrets` — the console's own admin password, the session secret, `api-keys.json` — and `iam-policies.json`, and a restore overwrites both wholesale. The `admin` tier's `backups:*` wildcard and API key levels both absorbed the new actions automatically, so admin could download an archive it created with a passphrase of its own choosing, and an existing key minted when `backups: write` meant `pg_dump` gained whole-host takeover on upgrade with no re-save and nothing to review.

### Also fixed

- **A quoted `.env` value no longer survives a restore as a bare one.** `config_value` strips the quotes to read a value and the writer put them back unquoted, so `DUNE_DB_PASSWORD="two words"` came back as an empty key followed by a stray command, and the password fell through to its default. Two existing suites already assert that this shape survives.
- **`battlegroup-restore-point.env` and `SERVER_IP` stay this host's** across the wholesale `runtime/generated/` extract. The rollback point otherwise ended up describing the *source* host, and `SERVER_IP`'s careful preservation in `.env` was silently overruled by the archive's `battlegroup.env`, which every consumer sources second.
- **The download that streams every credential on the host is rate limited**, like the other system-backup routes.
- **A stopped Battlegroup no longer reads as a starting one.** `isGameStackDown()` decides on the seven game containers alone, and Home reports **Database Only** instead of a bare "Stopped" over a live database. This changes what the console reports generally, not only after a restore.
- **The deferred dimension reconcile outlives a console reload.** It was backgrounded inside the console container, which `dune console reload` destroys, so on a restore it was killed every time and the second Sietch of a multi-dimension map was never spawned.

### Worth review

- Each route carries its own write-shaped IAM action rather than inheriting the `/api/backups/` prefix bucket's `backups:read`. The download's explicit entry is what keeps an archive of every host credential off a read-only grant.
- The passphrase travels in the task's env — never argv, the payload, or the audit log.
- Restore replaces the database *before* `.env`, and never restarts the stack: a restored `.env` can change the admin password and DB credentials.
- Host-shaped `.env` keys are preserved across a restore, `DUNE_DB_PASSWORD` included — it belongs to this host's Postgres role, which nothing ever `ALTER`s.
- **Upstream's new world-partition reconcile runs outside `finish_asset_phase`.** That helper is shared with `install-assets`, which must not touch the database, and the reconcile writes to it — so it stays in the full-update path, after the helper, since it reads the partition catalog the helper regenerates.
- **The receipt binds the archive and the principal, but not the restore options as reported.** Both callers preview *before* the operator picks the identity and audit-log modes — the preview is what reveals that a choice is needed — so binding them strictly would refuse every legitimate apply. A mode the preview left empty may be answered at apply; changing one it already carried is refused.
- **The receipt is an intent confirmation, not an authorization control.** Preview and apply are the same action and the same credential, so it closes the "lost or mis-sent `apply` flag" hole and nothing wider. The IAM changes above are what reduce the blast radius.

### Testing

From CI on `a851de34`, rebased onto upstream `v1.4.25`, all 11 checks green: `console/api` 2220 pass / 0 fail / 3 skipped (2223), `console/web` 1033/1033 across 92 files, and every shell suite in `runtime-script-unit` passing — including `tests/db-system-backup-test.sh` at 55 cases.

Safety properties are mutation-verified rather than merely covered: each guard was reverted individually and fails exactly the case that owns it.

Verified on a real host, not only in tests:

- **A full restore driven through the first-run wizard in a browser**, on a host reset to a genuinely fresh state — game files installed, passphrase verified, database/config/secrets replaced, Battlegroup started, console reloaded. 196 tables and the whole game stack came back up, including the deferred reconcile and the second Sietch of a multi-dimension map.
- **The preview gate end to end over HTTP**: preview → apply succeeds, a replayed apply is refused with 409, and a fresh preview re-authorizes it.
- **The principal binding, with real credentials.** A second logged-in browser session is refused 409 on the first session's preview. An API key stored as `backups: write` is refused 403. A key granted `backups:restore-system` explicitly previews and applies. The browser session is refused 409 on the key's preview.
- **Host-shaped preservation against an archive that genuinely differed**: a quoted `.env` value, `SERVER_IP`, `SERVER_IP_MODE` and the rollback point were all kept as this host's while the archive carried other values.
- **The wizard's retry path**: a wrong passphrase stops the restore, Retry Restore appears (and survives stepping back and forward), and retrying completes the restore.
- **The archive-digest guard**, refusing a mismatch before decryption and accepting a match.

**Not verified live:** the wizard's resume-after-reload path, which is covered by mutation-verified jsdom tests only. On a freshly reset host the apply finishes in seconds, so there was no window to reload into.


drkshrk:
tested deploy and restore.

Red-Blink:
The required restore preview is currently enforced only by the web interface. A caller can send `apply: true` directly to the restore API, which causes the runner to omit `--dry-run` without verifying that a successful preview occurred.

Because this operation replaces configuration, secrets, and the database, the API must enforce the safety gate. Please require a short-lived preview receipt bound to the authenticated session, archive identity/hash, and selected restore options.

Applying should be rejected if no matching preview succeeded or if the archive changed afterward.

Please also add integration tests covering direct apply rejection, archive replacement after preview, and a successful preview followed by apply.

drkshrk:
Thanks — confirmed, and it was worse than reported: `runner.js` sets `DUNE_DB_ASSUME_YES`, so `db.sh`'s typed confirmation didn't gate the API path either.

Fixed as asked: a successful dry run is recorded against the calling session or API key, the archive, and its sha256; an apply without a matching, unexpired receipt gets 409 (TTL via `ADMIN_RESTORE_PREVIEW_TTL_MS`, default 15 min). The three integration tests you asked for are in, plus failed-preview, cross-principal and expiry cases.

One deliberate deviation: the receipt doesn't bind the restore options strictly. Both callers preview *before* the operator picks the identity and audit-log modes — the preview is what reveals a choice is needed — so strict binding would refuse every legitimate apply. A mode the preview left empty may be set at apply; changing one it already carried is refused.

Two things this turned up, detailed in the PR description:

- **TOCTOU:** the route hashed the archive seconds before `db.sh` opened it, and import can rename another archive onto that name. `restore_system` now hashes and decrypts its own private copy.
- **Escalation:** the `admin` tier's `backups:*` and existing `backups: write` API keys silently gained the new system-backup actions. `download-system`, `import-system` and `restore-system` are now denied to both (keys can still name them explicitly) — a behaviour change for existing deployments.

Verified live on a real host, including the cross-session and API-key refusals.

PR #213, opened 2026-09-07
Stop reporting a lone Postgres as a starting Battlegroup

drkshrk:
On a host where Postgres is up but the Battlegroup is not, the console reported the Battlegroup as **starting** — permanently — and disabled Start. There was then no way to start the server from the console, and every appearance that something was wrong.

`isHomeBootStarting` ends in `return coreStartupContainerUp || …`, and that pattern lists `dune-postgres`. A lone database made it true, which forces `stopped` to false and `starting` to true in `getHomeServerState` regardless of everything else.

`getHomeServerState` could not correct it either: its only route to `stopped` required *every* container to be missing, including `dune-orchestrator` and `dune-coriolis-coordinator` — which the console keeps running by design, so on a console-managed host that is close to unreachable.

### Changes

- `isGameStackDown()` decides on the seven game containers alone (`rmq-admin`, `rmq-game`, `text-router`, `director`, `server-gateway`, `server-survival-1`, `server-overmap`). Both functions consult it. It returns false on a partial container list rather than guessing.
- Home reports **"Database Only"** rather than a bare "Stopped" when the database is up and the game stack is not. That state is one the console creates on purpose — it starts Postgres itself for backups and restores and leaves it up — so flattening it into "Stopped" hides a running database, and calling it "starting" was the bug.

### How to reproduce on a real host

Any state where Postgres is up and the game stack is not: after a backup or restore, or after `runtime/scripts/start-postgres.sh` on its own. Home shows `Overall: ISSUE` with Start greyed out.

### Testing

`console/web` 909/909, typecheck and build clean. The new test uses a real `dune status` capture verbatim rather than a sketch, so it pins the actual case; removing either call site of `isGameStackDown` fails exactly the case that owns it.

### DIFF
--- a/console/web/src/features/server/ServerPanels.battlegroupState.test.ts
+++ b/console/web/src/features/server/ServerPanels.battlegroupState.test.ts
@@ -1,0 +1,90 @@
+import { describe, expect, it } from "vitest";
+import { getHomeServerState, isGameStackDown } from "./ServerPanels";
+
+// Verbatim from a host after a system restore: the console started Postgres
+// itself, and nothing else in the Battlegroup is up. Every field the state
+// functions read is present, so this pins the real case rather than a sketch.
+const restoredHostStatus = [
+  "=== Dune status ===",
+  "Overall:     ISSUE",
+  "Title:       SteelHeart",
+  "Battlegroup: sh-372e084875362ee5-krlquy",
+  "",
+  "=== Containers ===",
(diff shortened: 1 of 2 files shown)


drkshrk:
Folded into #208, which carries the same commit — that PR already touches the restore flow this fixes, so reviewing them together is simpler than in isolation.

Issue #11, opened 2026-06-17
IP address not updating after 24h disconnect since latest update (players unable to connect)


Selyria:
Since the latest update, the IP address is no longer being detected correctly. Because the server fails to update and change the IP address, players are unable to connect to the Dune server.

I have attached a few screenshots of the issue. I have a daily 24-hour disconnect from my ISP and get assigned a new IP address every time. Previously, a simple restart always fixed it without any issues, but since the latest version, it somehow doesn't work anymore.

<img width="1895" height="907" alt="Image" src="https://github.com/user-attachments/assets/aa14f6cc-3016-4655-8773-074351b6959b" />

<img width="1093" height="144" alt="Image" src="https://github.com/user-attachments/assets/e166832b-abed-4e33-9e1d-bb7f538d0513" />

I tried

Restarting the server.
Restarting the entire server.
Restarting the home server.

None of these worked.

Red-Blink:
You are using an old version. The latest version as of now if v1.3.15. I am not sure if you are on discord, but it would be better to join so that I can help you better.

Selyria:
okay

Red-Blink:
That clearly shows why, your local linux permissions is a bit off, do this:

```bash
cd ~/dune-awakening-selfhost-docker

sudo chown -R "$USER:$USER" ~/dune-awakening-selfhost-docker
chmod 644 .env
```
Then retry the IP reconcile check:
```bash
echo "=== configured IP ==="
grep -E '^(SERVER_IP|SERVER_IP_MODE|SERVER_BIND_IP)=' .env

echo "=== current public IP ==="
curl -4 ifconfig.me
echo

echo "=== reconcile attempt ==="
runtime/scripts/network-addresses.sh reconcile

echo "=== configured IP after reconcile ==="
grep -E '^(SERVER_IP|SERVER_IP_MODE|SERVER_BIND_IP)=' .env
```
If `.env` still has the old IP after that, then we can debug the reconcile logic.

Right now, though, the script is failing before it can do anything because `.env` is not readable by the user running the command.

Selyria:
I just wanted to confirm that the fix works—thanks for that. 1.3.16

PR #215, opened 2026-09-11
console: Discord Bot settings -- hosted-bot auto-invite wizard + self-hosted adapter management (stacked on #201/#202)

yacketrj:
# Discord Bot Settings — self-service adapter management + hosted-bot auto-invite wizard

> [!IMPORTANT]
> **Do not merge this PR before #201 and #202 are merged.** This branch is built on `tier1-upstream`, which already includes both PRs' commits — its diff against `main` currently contains #201/#202's changes as well as this PR's own. GitHub does not enforce merge order between independent PRs even when one is stacked on another (confirmed: all three currently show `mergeStateStatus: CLEAN` against `main` independently) — merging this PR first would merge #201/#202's changes prematurely, without their own review having concluded. Correct order: **#201 → #202 → this PR**, rebasing/re-diffing this PR after each merges so its diff against `main` narrows down to only its own changes.

**Stacked on #201 (Tier 3 TOTP) and #202 (Tier 1 Discord OAuth sign-in).** This branch is built on `tier1-upstream`, which already includes both.

## Executive Summary

Today, upstream's only Discord-adapter-settings surface is a single read-only `discordAdapterEnabled(config)` boolean — driven by a static env var, restart-to-apply, zero runtime API, zero settings UI. An operator who wants the console's Discord bot feature can only configure it by hand-editing `.env` and restarting the whole stack.

This PR adds the entire self-service **Settings → Discord Bot** experience: enable/disable, adapter-token generation/regeneration, role-ID management, and — the reason this exists at all — a **one-click hosted-bot auto-invite flow**. An operator who doesn't want to run and maintain their own Discord bot process can instead invite a shared, hosted bot (this fork's own instance, "Sahir Venn," to a Discord Application ID that's fully operator-overridable — see `autoInviteDiscordClientId`), have it join their server with one click, confirm ownership, and be connected — no manual token copying, no second Discord Application to register, no CLI.

This is fully additive to the read-only boolean upstream has today: an operator who never opens Settings → Discord Bot sees no behavior change, and `DUNE_DISCORD_ADAPTER_ENABLED` continues to work exactly as it does now for anyone who prefers to keep hand-editing `.env`.

---

## What This PR Accomplishes

- **New Settings → Discord Bot section** (`console/web/src/features/settings/DiscordBotSection.tsx`): a 3-step guided flow (choose hosted vs. self-hosted → connect/configure → done), plus an ongoing-management view for an already-enabled adapter (role IDs, token regeneration, disable).
- **Hosted-bot auto-invite** (`console/api/src/integrations/discord/autoInvite.js`, `hostedBotOAuth.js`): the operator clicks "Add bot to Discord," Discord's own native bot-install consent screen runs, the operator picks their guild, Discord OAuth (a *separate*, dedicated Discord Application — never the console-sign-in one) verifies real guild ownership, and the console forwards the (already-generated) adapter token to the hosted bot backend server-to-server. The operator never sees or copies the token themselves for this path.
- **Self-hosted path unchanged in spirit, generated-not-typed in practice**: an operator running their own bot instance still gets a real adapter token, minted and shown once by the console (same one-time-reveal convention as every other secret this project generates), with the same enable/regenerate/disable actions available.
- **New backend routes** (`server.js`): `GET/POST /api/settings/discord-bot{,/enable,/role-ids,/regenerate-token,/disable,/restart,/choice}`, plus the hosted-bot OAuth start/callback/register/confirmation-status routes under `/api/integrations/discord/hosted-bot/`.
- **`age`-based encryption at rest** for the two new secrets this feature introduces (the hosted-bot OAuth Discord Application's client secret, and the adapter token) — wired into this fork's existing Stage 1/2 age-secrets migration tooling (`runtime/scripts/secrets-cli.sh`, `docs/security/age-secrets.md`), not a new, parallel secrets mechanism.
- **Generic, overridable Discord Application ID** for the auto-invite flow (`AUTO_INVITE_DISCORD_CLIENT_ID`, `config.js`) — not hardcoded to this fork's own hosted-bot identity, so a fork running a different hosted-bot backend can point this flow at their own Discord Application. (One real reachability gap in this — the env var isn't yet wired through `docker-compose.web.yml` for a real containerized deployment — is filed as [#915](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/915) rather than guessed at in this PR.)
- **The old, independent-Discord-Application manual-connect path is kept, not removed** — collapsed behind an "Advanced: connect manually instead" disclosure once a guild is already connected via auto-invite. A decision to eventually remove it outright (rather than keep it as a permanent fallback) was made explicitly during design, but deliberately **not bundled into this PR** — it's gated on the new flow being confirmed working end-to-end against a real deployment first, which is documented, not silently dropped.

---

### The problem, in the operator's own words

From live UAT of an earlier iteration of this same feature, before the auto-invite flow existed:

> "why are we asking for Redirect URI during the Hosted Bot setup? We are hosting the bot, we know the redirect URL."

The original hosted-bot flow (still present in this PR, behind "Advanced: connect manually instead") required the operator to register their own, independent Discord Application — Client ID, Client Secret, Redirect URI, all filled in by hand — purely to prove they owned a Discord server, even though the bot itself is hosted centrally and the redirect URL is always the same, known value. This is real, unnecessary friction for the common case.

### Why an OAuth-registration flow at all, not just a text field

A closely related problem this PR also fixes: `mentat` (the hosted, multi-tenant bot backend) is a single running process serving many operators' consoles. Something has to register `{guildId → consoleUrl, adapterToken}` in its database. The **previous** way this happened was a fully separate onboarding flow: the bot DMs the inviter a link to a standalone setup portal, hosted on a different domain, where the operator manually pastes in a Console URL, an Adapter Token (generated by running `openssl rand -hex 32` by hand and copying it back), and role IDs.

This had a real, previously-shipped-and-reverted failure mode (`mentat#194`): a "Generate" button on that old portal minted a **browser-side** random token that Core's console could never actually recognize (Core only ever validates against whatever is already in its own token file) — operators completed "successful" setup and then got a silent, unexplained 401 on every bot command. The fix at the time (`mentat#214`) was manual `openssl` instructions — a real regression in usability, adopted specifically to avoid repeating `#194`'s mistake.

This PR's design avoids `#194`'s root cause a different way: it **never manufactures a new token value at all**. The adapter token is generated once, by Core, the same way it already is for the self-hosted path (`enableDiscordBotAdapter()`/`regenerateDiscordBotToken()` in `adapterSettings.js`) — this flow's only job is to read that same, already-generated token and hand the exact same bytes to the hosted bot's registration API, server-to-server. There is no independent token-generation code path anywhere in this design. Guild ownership is proven the same way: a live Discord OAuth token is forwarded to the hosted bot backend, which makes its own independent call to Discord's API with it — the backend never trusts "Core says this operator owns guild X" on Core's own say-so.

### Why the console-initiated flow instead of keeping the separate setup portal

Moving registration into Core's own Settings page — the one place an operator already goes to manage everything else about their install — means: no context-switch to a different site, a real path back in to reconfigure or connect a second guild (the old portal only worked from a fresh Discord DM), and one fewer system for an operator to have degraded understanding of when something goes wrong.

---

## Testing

Real, current output from this exact branch (`upstream-pr/wizard-hosted-bot-settings`, commit `e14c4cad`):

```
$ cd console/api && npm test
...
# tests 2633
# suites 6
# pass 2630
# fail 1
# cancelled 0
# skipped 2
# todo 0
```

The 1 failure (`root scheduled backups remain private and can be read by the Console user`, `test/marketBotBackupPrune.test.js`) is a pre-existing, unrelated test that depends on dropping to a non-root user inside the test sandbox — confirmed to fail identically against `tier1-upstream` alone, unmodified by this branch, in this same CI-less local environment. Not something this PR introduced or can fix.

```
$ cd console/web && npm test
...
 Test Files  103 passed (103)
      Tests  1103 passed (1103)

$ cd console/web && npx tsc --noEmit
(clean, no output)
```

This diff also went through 3 rounds of dedicated regression-guard verification during a Layer 3 integration audit (see below) — for each of 6 CRITICAL/HIGH findings, the fix was reverted, the corresponding new test was confirmed to genuinely fail with the real, exact error, then the fix was restored and reconfirmed green. This is not a claim; see the Eight Hats Audit Summary section for what each of those was.

### Security checks

```
$ gitleaks detect --source . --log-opts="origin/tier1-upstream..HEAD" --config .gitleaks.toml
17 commits scanned.
scanned ~682496 bytes (682.50 KB)
no leaks found

$ semgrep --config p/default --config p/secrets --quiet --error <56 changed files>
(clean, exit 0)

$ shellcheck -S warning <11 changed shell scripts>
(clean, exit 0)
```

---

## Fresh Install

No change to fresh-install behavior for an operator who never opens Settings → Discord Bot. `DUNE_DISCORD_ADAPTER_ENABLED` continues to gate the adapter exactly as it does on real upstream `main` today. Two kinds of new env vars, correctly distinguished:

- **Operator-facing, documented in `.env.example`**: `DISCORD_HOSTED_BOT_OAUTH_CLIENT_ID`/`DISCORD_HOSTED_BOT_OAUTH_REDIRECT_URI` (the client secret itself, like every other secret in this project, lives in `runtime/secrets/`, never `.env.example`). All default to empty (feature inactive).
- **Console-managed internal state, never meant to be hand-edited**: `DUNE_DISCORD_ADAPTER_DEPLOYMENT_CHOICE`, `DUNE_DISCORD_HOSTED_BOT_CONNECTED_GUILD_ID`/`_NAME` are written by the console itself (`adapterSettings.js`'s `MANAGED_ENV_KEYS`) as bookkeeping, the same way this project already manages several other `.env` keys no operator is expected to type by hand — deliberately absent from `.env.example` for that reason, not an oversight.
- **Override-only, code-level defaults**: `AUTO_INVITE_DISCORD_CLIENT_ID`/`_REDIRECT_URI`, `MENTAT_LINK_CONFIRMATION_STATUS_URL` (`config.js`) default to this fork's own hosted-bot identity/backend and are override escape hatches for a differently-hosted deployment, not vars a typical operator sets — also not in `.env.example`. One real gap in how reachable this override actually is for a real deployment is filed as [#915](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/915).

None of the above are secrets in the sensitive sense (a Discord `client_id`/URL is not confidential), so their absence from `.env.example` is a documentation-scope choice, not a security one.

## Migration / Upgrade Path

- **An operator already using `DUNE_DISCORD_ADAPTER_ENABLED` + manually-set role-ID env vars sees zero disruption.** `discordRoleMappingFromEnv()` reads the same env vars as before; nothing about the existing manual configuration path is removed or renamed.
- **`DISCORD_OBSERVER_ROLE_IDS` → `DISCORD_PLAYER_ROLE_IDS` rename** (a pre-existing rename this branch's base, `tier1-upstream`, already carries — not introduced by this PR) keeps a documented legacy-fallback read: an operator who already set the old name keeps working across the update with no manual step, and the new field name only takes precedence when explicitly set (including explicitly cleared to empty — see the `!== undefined` distinction in `adapter.js`, a real Layer 3 finding fixed during the port of this branch, not present in the original implementation).
- **No schema change, no required new environment variable, no changed default** for anyone not touching this feature.
- **The two new secrets this PR introduces** (hosted-bot OAuth client secret, adapter token) default to plaintext-on-disk (`runtime/secrets/*.txt`, mode 600) — the same convention every other secret in this project already uses on a fresh install. Migrating either to `age`-based encryption at rest is opt-in, documented in `docs/security/age-secrets.md`, and follows the exact same `dune secrets migrate/verify/cleanup-legacy` flow as this fork's existing Stage 1/2 secrets.

## Break/Fix — What Happens Mid-Run

- **A settings save (.env write) always succeeds or fails atomically and independently of the container-apply step.** `updateEnvFileValues()` (this branch's base) uses a real write-queue serialization and atomic temp-file-then-rename write — confirmed directly during this branch's own port work (a real API mismatch between this feature's original `.env` writer and this base branch's hardened, queued/async version was found and fixed here, not silently left broken).
- **The container-recreate step that applies a saved setting can fail.** `recreate_discord_adapter_env()` (`self-update.sh`) reports a clear `failed` status with a log-file pointer on a `docker compose up` failure. **One real gap found in this PR's own Layer 3 audit is not yet fixed**: that function's `docker rm -f` runs before the recreate, with no rollback if the subsequent `up` fails — filed as [#914](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/914) rather than guessed at, since the right fix depends on why the manual `rm -f` exists at all, which isn't documented anywhere this audit could find.
- **A hosted-bot registration call that fails partway** (network failure, hosted-bot backend down) is retried exactly once (`httpWithRetry.js`) and surfaces a clear, actionable error to the operator — never silently retried more than once, and never left in a partially-registered state (registration is a single atomic call on the hosted-bot backend's side; Core only ever stages "pending" state locally until a real confirmed response comes back).
- **An operator who abandons the hosted-bot flow partway through** (closes the tab after Discord OAuth but before confirming) leaves no persisted state on Core's side — the pending confirmation record expires on its own TTL, and nothing was written to `.env` until the operator explicitly confirms.

## Break/Fix — What Happens if a Credential Is Lost

- **Hosted-bot OAuth Discord Application client secret lost or leaked**: reset it in the Discord Developer Portal and replace `runtime/secrets/discord-hosted-bot-oauth-client-secret.txt` (mode 600) with the new value, or use Settings → Discord Bot → Advanced → change application credentials. If migrated to `age` encryption, `dune secrets migrate discord-hosted-bot-oauth-client-secret` re-encrypts the new value the same way. No lockout of anything else — this secret only gates the auto-invite/manual-connect OAuth round-trip, not console sign-in or the adapter's own token-based bot auth.
- **Adapter token lost, leaked, or simply needs rotating**: Settings → Discord Bot → Regenerate Token mints a fresh one immediately and clears any legacy direct-env-var value that might otherwise silently keep authenticating with the old one (a real, audited finding from this feature's original development — see the Eight Hats summary below). The old token stops working the moment the container recreate completes.
- **Both are recoverable by the operator at any time** — unlike PR #159's system-backup passphrase, nothing in this PR introduces a credential with no recovery path. The only "loss" scenario that requires operator action is re-registering with the hosted bot backend, which is a normal, supported, idempotent operation (re-running the auto-invite flow), not a special recovery procedure.

## Rollback Safety

This feature can be fully disabled without data loss: Settings → Discord Bot → Disable Discord Bot Integration resets every managed `.env` key this feature owns back to unset (`disableDiscordBotAdapter()` in `adapterSettings.js`), deletes the token file, and clears any hosted-bot connection record — an operator who disables and later re-enables goes through the wizard from a genuinely clean first step, matching what a fresh install looks like. No irreversible action anywhere in this feature's disable path.

---

## Eight Hats Audit Summary

This feature was built and audited on this fork's own `main` branch over several weeks before this port to `tier1-upstream` for this upstream PR — the layered audit discipline below reflects that full history, not just this port.

**Layer 1 (design)** — two dedicated design documents, each independently audited before implementation began:
- `hosted-bot-oauth-registration-l1-design-2026-09-09.md` — the original console-initiated registration design. Audit findings register + STRIDE table in its own §10.
- `hosted-bot-auto-invite-and-role-picker-l1-design-2026-09-10.md` — the auto-invite flow that superseded the original design's manual-form UX (tracking issue #832; findings filed as issues #833-#852 across 3 audit rounds, plus #876 in a 4th round after live UAT surfaced a real gap in how Core learns about a completed hosted-side confirmation). All rounds' CRITICAL/HIGH findings resolved before implementation proceeded past each gate.

**Layer 2 (per-feature implementation)** — dispatched per PR as each piece shipped (PRs #748, #801, #866, #868, #871-873, #891, #893-895, #897-898). Representative CRITICAL/HIGH findings actually caught and fixed, not just claimed:
- **#859 (CRITICAL)**: an admin-tier console operator could substitute the hosted-bot Discord Application's own Client ID/Secret/Redirect URI, then have the owner unknowingly complete a real Discord consent flow against the hijacked config, handing their actual Discord identity/access token to whoever substituted it. Fixed by making that action owner-only (`settings:discord-bot-oauth-config`/`-oauth-secret`), matching every other credential-replacement action in this feature.
- **#870 (CRITICAL)**: a guild name containing `$(...)`/backtick shell-metacharacters, written into `.env` and later `. sourced` by this project's own startup scripts, executed as a real shell command with the script's own privileges — found by automated review *after* the PR that introduced it had already merged. Fixed by stripping every shell metacharacter (not just the original, incomplete control-character/`=` blocklist) from guild names before they're ever written.
- **A real, live UAT finding, not a hypothetical**: "I see no path to remove the bot" — this feature originally shipped an Enable/Save/Regenerate surface with no way back to "never configured." Added the Disable action described above.
- Several audit findings on token-precedence/staleness correctness (a regenerated token not becoming authoritative when a legacy direct-env-var value or legacy token-file-path variable was still set) — each fixed and covered by a dedicated regression test at the time.

**Layer 3 (integration, this PR specifically)** — dispatched against the full 17-commit port-to-`tier1-upstream` diff via a local `/code-review high` pass (upstream's own `/code-review ultra` failed to start twice on this cross-fork PR — a real product bug, reported separately; not something this PR's own audit could route around). Every finding was independently re-verified against real code/tests before being treated as confirmed, per this project's own "an internal audit's claim is a hypothesis until independently confirmed" discipline:

| Severity | Finding | Status |
|---|---|---|
| CRITICAL | `buildDuneArgs()` had no case for `"discordAdapterApply"` — every settings-apply task unconditionally failed, silently (the `.env` write succeeds separately, masking it live) | Fixed |
| CRITICAL | The hosted-bot OAuth state cookie's `Secure` attribute was conditional on a flag that's `false` by default — `SameSite=None` without `Secure` is rejected outright by every modern browser, so the cookie never survived the Discord redirect on any default install | Fixed |
| HIGH | A 5xx-then-throws sequence in the hosted-bot registration retry helper made a 3rd real network call instead of the promised single retry — a duplicate-side-effect risk for a non-idempotent POST | Fixed |
| HIGH | The Enabled-phase choice toggle revealed the "Connect to hosted bot" UI before the choice was actually persisted server-side — clicking through in that window hit a real 403 | Fixed |
| HIGH | A shell-side adapter-health-check token resolver used the opposite credential precedence from the real server, risking a false "unhealthy" report | Fixed |
| HIGH | A Discord API call during the hosted-bot OAuth callback had no timeout at all, unlike its sibling helper — a real hang-indefinitely risk on the auth path | Fixed |
| HIGH | `recreate_discord_adapter_env()`'s `docker rm -f` before `up --force-recreate` has no rollback if the recreate fails | Filed: [#914](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/914) |
| MEDIUM | `AUTO_INVITE_DISCORD_CLIENT_ID` (+2 siblings) can't actually be overridden by a real docker-compose deployment, despite that being the stated purpose of genericizing it | Filed: [#915](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/915) |
| LOW | A secrets-helper permission-drift edge case is silently indistinguishable from "never configured" | Filed: [#916](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/916) |
| LOW | A snowflake-regex inconsistency and one confirmed-dead-code duplicate of an already-fixed revocation-bypass bug | Filed: [#917](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/917) |

Every finding above marked "Fixed" was verified, not assumed: the fix was reverted, the corresponding regression test was confirmed to genuinely fail with the real error, then the fix was restored and reconfirmed passing. Full findings register and STRIDE table posted as a PR comment.

**Real upstream CI caught one more thing none of the above did**: 8 shell scripts in this branch's diff had lost their executable bit somewhere in this branch's merge history (a tooling artifact — `git diff --stat` before that fix showed zero content changes, pure mode bits). 3 of Red-Blink's own CI jobs failed identically on it; fixed in a follow-up commit, all 9 checks now pass.

### STRIDE mapping (Layer 3)

| Category | Finding(s) | Status |
|---|---|---|
| Spoofing | N/A this layer | — |
| Tampering | Duplicate-network-call finding (unintended duplicate server-side write); latent risk in the dead-code duplicate if ever wired up (stale role mapping gates real bot capabilities) | Fixed / Filed (#917) |
| Repudiation | N/A this layer | — |
| Information Disclosure | N/A — the cookie finding is a functional break (never sticks), not a disclosure of it | — |
| Denial of Service | `discordAdapterApply` always-fails bug, the Secure-cookie-always-fails bug, the unbounded Discord call, the no-rollback container recreate | Fixed / Fixed / Fixed / Filed (#914) |
| Elevation of Privilege | Latent role-mapping angle on the dead-code duplicate finding | Filed (#917) |

---

## Files Changed

56 files, +11,573/-58. Notable groupings:
- **New backend files**: `console/api/src/integrations/discord/{adapterSettings,autoInvite,hostedBotOAuth}.js`, `console/api/src/services/httpWithRetry.js`, and their test files.
- **New frontend files**: `console/web/src/features/settings/DiscordBotSection.tsx` (+test), `console/web/src/api/{discordAdapterSettings,discordHostedBotApi}.ts` (+test).
- **Modified shared files**: `server.js`, `config.js`, `actions.js`, `policy.js`, `tasks.js`, `runner.js` (new routes/actions/task-dispatch case), `adapter.js` (role-ID field precedence fix), `oauth.js` (shared `constantTimeStringEqual` export).
- **Modified infra**: `docker-compose.web.yml`, `.env.example`, `.github/workflows/ci.yml`, `runtime/scripts/{console.sh,secrets-cli.sh,self-update.sh}`, new `runtime/scripts/lib/console-secrets-env.sh`, 7 new `runtime/tests/*.sh`.
- **Docs**: `docs/console/API-REFERENCE.md`, `docs/integrations/discord-control-bot/setup-guide.md`, `docs/security/age-secrets.md` updated; new `docs/security/secrets-management.md`.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01M1EJFaWjE6V81HgKowD7Lv


yacketrj:
## Layer 3 (integration) Eight-Hat audit — findings register

Upstream's own `/code-review ultra 215 --post` failed to start twice with an identical error (cross-fork PR resolution bug, reported separately as product feedback). Ran a local `/code-review high` pass against the full branch diff instead, then independently verified every finding against real code/tests before treating it as confirmed — per this project's own Requirement 20 discipline ("an internal audit's claim is a hypothesis, not a fact, until independently confirmed").

**8 of 10 ranked findings confirmed real and fixed** (commit `ea513ac5`), each with a regression test that was verified to actually fail before the fix and pass after. **1 finding was already resolved** by a later commit already on this branch. **2 findings need maintainer/design judgment** and are filed as issues rather than guessed at.

### Findings

| # | Severity | File:Line | Summary | Status |
|---|---|---|---|---|
| 1 | **CRITICAL** | `console/api/src/runner.js` | `buildDuneArgs()` had no case for `"discordAdapterApply"` — every Discord Bot settings enable/role-ids/restart task unconditionally failed with `"Unsupported operation"`, silently (the `.env` write happens separately and synchronously, masking the failure live) | ✅ Fixed |
| 2 | **CRITICAL** | `console/api/src/integrations/discord/hostedBotOAuth.js` | `hostedBotOAuthStateCookie()` gated `Secure` on `config.secureCookies`, false by default (`docker-compose.web.yml`'s `ADMIN_SECURE_COOKIES:-0`) — `SameSite=None` without `Secure` is rejected outright by every modern browser, so this cookie never survived the Discord redirect on any default install | ✅ Fixed |
| 3 | HIGH | `console/api/src/services/httpWithRetry.js` | The 5xx-retry call sat inside the same `try` whose `catch` performs its own retry — a 5xx followed by the retry itself throwing made a 3rd real network call instead of the promised single retry (duplicate-side-effect risk for a non-idempotent POST) | ✅ Fixed |
| 4 | HIGH | `console/web/.../DiscordBotSection.tsx` | The Enabled-phase choice toggle only wrote to local state/localStorage; the "Connect to hosted bot" UI it reveals immediately depends on `deploymentChoice` being persisted server-side (only Save Role IDs did that) — clicking through in that window hit a real 403 | ✅ Fixed |
| 5 | HIGH | `runtime/scripts/self-update.sh` | `resolve_discord_adapter_token()` checked the token file before the direct env var — the opposite of `readDiscordBotApiToken()`'s (routes.js) real, current precedence; a stale comment's premise that routes.js was "also" fixed was never true | ✅ Fixed |
| 6 | HIGH | `hostedBotOAuth.js` `discordJsonRequest()` | No timeout at all, unlike `oauth.js`'s near-identical helper (5s bound, added specifically to prevent a Discord brownout hanging the request indefinitely — DoS on the auth path) | ✅ Fixed |
| 7 | LOW | `console/web/.../SettingsPanel.tsx` | Stale comment describing a since-removed state-based ordering swap | ✅ Already resolved (commit `812babd5`, earlier on this branch) |
| 8 | HIGH | `self-update.sh` `recreate_discord_adapter_env()` | Unconditional `docker rm -f` before `up --force-recreate`, no rollback if the recreate fails — a routine settings save could leave the console down with no auto-recovery | 📋 Filed: [#914](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/914) |
| 9 | MEDIUM | `config.js` / `docker-compose.web.yml` | `AUTO_INVITE_DISCORD_CLIENT_ID` (+2 siblings) can't actually be overridden by a real deployment — contradicts the stated upstream-facing goal of letting a self-hoster point the flow at their own backend | 📋 Filed: [#915](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/915) |
| — | LOW | `runtime/scripts/lib/secrets.sh` | Permission-drift on a secret file is silently indistinguishable from "never configured" | 📋 Filed: [#916](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/916) |
| — | LOW | `adapterSettings.js` / `discordAdapter.js` | Snowflake regex inconsistency; dead-code duplicate of a role-ID revocation-bypass fix (confirmed unreferenced by any live route) | 📋 Filed: [#917](https://github.com/Project-Arrakis/dune-awakening-selfhost-docker/issues/917) |

**Bonus catch during verification**: `test/composeEnvPassthrough.test.js`'s pass-through check was too narrow (only recognized `${KEY:-`, not `${KEY-}` — the latter is what an earlier fix in this same branch legitimately uses for `DISCORD_PLAYER_ROLE_IDS`, to distinguish "explicitly cleared" from "unset"). Fixed in the same commit.

### STRIDE mapping

| STRIDE category | Finding(s) | Severity | Status |
|---|---|---|---|
| Spoofing | N/A — no finding in this layer involved impersonating an identity | — | — |
| Tampering | #3 (httpWithRetry duplicate call → unintended duplicate server-side write); latent risk in the #11/dead-code finding if `discordAdapter.js`'s duplicate ever gets wired up (a stale role mapping gates real bot capabilities) | HIGH / LOW (latent) | Fixed / Filed (#917) |
| Repudiation | N/A | — | — |
| Information Disclosure | N/A — #2's cookie gap is a functional break (cookie never sticks), not a disclosure of it | — | — |
| Denial of Service | #1 (task always fails, feature never applies), #2 (cookie never survives the redirect, feature always fails), #6 (unbounded Discord call can hang the auth path indefinitely), #8 (no rollback → console stays down) | CRITICAL / CRITICAL / HIGH / HIGH | Fixed / Fixed / Fixed / Filed (#914) |
| Elevation of Privilege | N/A directly; see Tampering row for the latent role-mapping angle on #11 | — | — |

Findings #4, #5, #9, and the `secrets.sh`/regex items (#916/#917 body) don't map cleanly to a STRIDE category — they're functional-correctness, portability, and observability gaps rather than security-boundary issues, consistent with this project's own guidance that not every finding needs a STRIDE label.

### Verification method

Every "Fixed" row above was independently confirmed by reverting the fix, re-running the exact regression test added for it, observing the real failure (not a hypothetical), then restoring the fix and confirming green again. Full suites after all fixes: `console/api` 2630/2631 (the 1 remaining failure is a pre-existing, unrelated sandbox-root limitation in `marketBotBackupPrune.test.js`, confirmed to fail identically on `tier1-upstream` alone in this environment), `console/web` 1103/1103, `tsc --noEmit` clean, gitleaks + semgrep clean on the diff.

This PR stays in draft — the full Requirement 19(g) body (executive summary, upgrade/break-fix behavior, real test/scan output, complete Eight-Hats summary) still needs to be written to replace this placeholder before it's marked ready.

yacketrj:
Rebased onto current \`upstream/main\` (35 commits behind, now current).

**Flagging a CI result I could not fully resolve**: \`runtime-script-unit\` failed on this push (\`Release gate\` failing too is just the expected downstream consequence). The job log shows no error/assertion output at all -- it ends abruptly, exit code 1, right as \`runtime/tests/test-deepdesert-layout-apply-safety.sh\` starts (the previous test, \`test-deepdesert-dual-selector-cleanup.sh\`, printed a clean PASS immediately before).

Ran that specific test locally 5 times in a row -- passes clean every time, no timing-sensitive constructs (\`sleep\`/\`timeout\`/background jobs) in its ~230 lines that would explain CI-only flakiness. I don't have permission to trigger a CI re-run on this repo to check reproducibility there (only a re-push does that, and I don't have a substantive follow-up change to push right now). Given a clean local repro and no obvious cause in the test's own content, this looks like CI-runner-side (resource contention, a runner hiccup), but I want to be upfront that I couldn't independently confirm that rather than assert it -- if it reproduces on a real re-run, that's a different, more serious finding worth investigating further.

yacketrj:
Update: branch is now current with upstream/main (0 commits behind) after merging in #201/#202. `runtime-script-unit` failed again with the identical signature reported earlier — abrupt `exit code 1` immediately after the "Dual Deep Desert selector cleanup" test passes, right as `test-deepdesert-layout-apply-safety.sh` would start, no error output. Still reproduces cleanly locally. This looks CI-runner-environment-specific rather than a code regression, but I can't independently confirm that without write access to re-run the job.

Red-Blink:
This PR is not safe to merge yet.

First, please rebase it after updated versions of `#201` and `#202` are rebased onto the current upstream `main`, then resolve the inherited conflicts across authorization, server, frontend, tests, and API-key documentation.

There is also a confirmed high-risk availability issue in `recreate_discord_adapter_env()`.

It currently runs:

```text
docker rm -f
```

before attempting:

```text
docker compose up
```

If the recreate fails, the previously working Console has already been removed and there is no rollback or recovery path.

Please change this flow so a failed Discord-settings apply either:

- Leaves the existing Console running, or
- Reliably restores the previous working Console if the recreate fails.

Please also add a regression test that verifies actual service continuity or recovery, rather than only checking that the status file reports `failed`.

The advertised hosted-bot overrides also need to work in a real Compose deployment.

The API reads:

- `AUTO_INVITE_DISCORD_CLIENT_ID`
- `AUTO_INVITE_DISCORD_REDIRECT_URI`
- Related hosted-service endpoint overrides

but these are not currently passed through `docker-compose.web.yml`.

Please either wire those values through Compose or remove the claim that operators can override them.

Finally, `runtime-script-unit` and the dependent release gate are currently failing. The reported failure occurred around:

```text
runtime/tests/test-deepdesert-layout-apply-safety.sh
```

It passed when I ran it locally against this exact head, so please investigate or rerun it after rebasing and ensure every required check is green before requesting another review.

yacketrj:
## Layer 3 (integration/pre-PR) Eight Hats audit — PR #215, 2026-09-24

Dispatched as 2 independent workers (QA+UI/UX; Security+Architect+GRC+DBA+Network+CloudSec) against this PR's own incremental diff on top of #202 (hosted-bot auto-invite wizard + self-hosted adapter management).

### HIGH finding — fixed in `36ba18d1`

**Security/Cloud Security hats:** `DISCORD_HOSTED_BOT_OAUTH_CLIENT_SECRET` had no `EnvManaged` guard, unlike its two siblings (`discordOAuthClientSecretEnvManaged`, `discordBotHandoffSecretEnvManaged`). Consequence: an operator setting this var directly got no startup exposure warning (Requirement 24's own convention, applied everywhere else, silently skipped here), and `POST /api/settings/discord-bot/oauth-secret` would write a new value to disk and report success while the env var kept winning after the next restart — a silent no-op, the exact bug class already found and fixed once for the sibling secret (`saveOAuthClientSecret()`).

**Fix:** added `discordHostedBotOAuthClientSecretEnvManaged`, the matching startup warning, and the same refuse-outright guard on the rotate route — mirrors `saveOAuthClientSecret()`'s established precedent exactly. Full suite re-verified: 2849/2852 api (1 known pre-existing sandbox-only failure + 1 confirmed-flaky concurrent-DB-contention test, both unrelated and non-regressions), 1398/1398 web.

### Process note (corrected during this audit)

The dispatched agent's first pass reported the branch as "170 commits behind `origin/tier1-upstream`" — independently re-verified directly (`git merge-base --is-ancestor`, `git rev-list --count`) and this was a bad ref comparison in that pass, not real staleness; the branch was correctly rebased. Noting here so this doesn't get treated as an open item.

### Other findings (non-blocking)

|---|---|---|---|
| Architect | MEDIUM | Rotate route has no length/entropy ceiling on the new secret, only a 20-char floor. | Cosmetic, consistent with the sibling secret's own validation depth — not a regression |
| GRC | LOW | Setup guide doesn't explicitly describe partial-failure/recovery UX for a mid-flow 502; the in-app error copy is itself the recovery instruction. | Non-blocking |
| QA | MEDIUM | `handleStartAutoInvite()`'s catch branch (429/502/500 from `/auto-invite/start`) has no test making the call actually reject — by-inspection looks correct (popup closed, error set, error banner not step-gated) but unverified by a real test. | Open, recommend adding before merge per Requirement 20's MEDIUM disposition |
| UI/UX | LOW | Top-level error banner has no `role="alert"`/`aria-live`, inconsistent with every other transient status message in this component (`role="status"`). | Non-blocking, one-line fix suggested (`role="alert"` on the error wrapper) |

### Verified clean (no finding)
SSRF surface (Core never fetches operator-supplied `consoleUrl` itself, only forwards it — mentat-link's own validation is out of this repo's scope), token custody (bot token never touches Core; the adapter token Core does hold follows the established `runtime/secrets/` 0600 convention, redacted in audit logs, returned to the frontend exactly once on mint), no schema/migration surface, no new inbound endpoints beyond the documented OAuth/auto-invite routes (same-origin, cookie-scoped), popup-blocked/closed/cross-origin-postMessage-spoofing/confirmation-poll-across-remount all already covered by existing code+tests from prior review rounds (extensive in-code comment trail of already-fixed findings, #860/#870/#876/#903).

### STRIDE

|---|---|---|---|
| Information Disclosure | Hosted-bot secret EnvManaged gap | HIGH | **Fixed** (`36ba18d1`) |
| Repudiation | Same finding (silent UI-rotate no-op — operator believes they rotated, they didn't) | HIGH | **Fixed** |
| Spoofing / Tampering / Denial of Service / Elevation of Privilege | No findings — CSRF/postMessage-origin/confirmation-cookie coverage already verified sound by prior review rounds, re-confirmed by this pass | — | — |

PR #216, opened 2026-09-14
Fix CHOAM ranged weapon category masks (#12)

jeffstokes72:
Icehunter/EDA CategoryMask() remapped melee under Weapons folder 0 at depth 3, but left each gun type as a depth-2 sibling (pistol=2 through lasgun=13). Funcom's prefix filter then put Maula pistols in the Ammunition folder and left Ranged Weapons empty. Nest those type codes under Ranged Weapons at depth 3, move ammunition to folder 2, and remap on seed/buyback/CSV load so stale Icehunter plans cannot reintroduce it.


Red-Blink:
Thanks for working on this! I found one issue that needs fixing before merge.

`normalizeExchangeCategory()` currently remaps every depth-2 **Weapons** category between codes `2–13`, regardless of item kind. This can incorrectly move a custom schematic from the **Unique Schematics** folder (`0x01030000`, depth 2) into **Ranged Weapons subtype 3**.

Please restrict the legacy ranged-category remapping to:

```text
kind === "equippable"
```

Ammunition can continue using its existing separate handling.

Please also add a regression test confirming that depth-2 schematic and other non-equippable rows remain unchanged.

Everything else looks good, and all automated and local tests passed.

